PR #16, opened 2023-10-19
TEST: nodejs-v18 build

anxolin:
# Description
Require NODEJS v18 (not v20)

The reason, v18 is the LTS and some platforms like vercel don't support above that version:
image

CTX
https://cowservices.slack.com/archives/C04N05LJK54/p1697734199857449

## Edit 1: First attempt
Unfortunatelly, you can run it (`yarn start`) but if you try to build you get this:
image

## Edit 2: Dokusaurus 2.0 only requires v16
I'm a bit unsure how this project was created. I see that latest version of docusaurus don't have this restriction. Also, i notice something really stange in the versions of Dokusaurus (it uses 0.0.0?? instead of 2.4.3??)

Versions I get in a fresh project created with the following command: 
* `npx create-docusaurus@latest my-website classic --typescript`

image

See https://docusaurus.io/docs

This project uses 0.0.0??
image


mfw78:
I configured docusaurus using the `canary` tag in order to make use of strictly typed configuration / sidebars. This was working well. There is some very odd behaviour with the `swagger-ui-react`, which itself is a React 17 component and for some reason fails compilation with node 18, but passes with node 20.

It would be great if we can ensure that strict typing remains available, and work out whatever this weird `swagger-ui-react` issue with node 18 is.

mfw78:
Fixed in #22

PR #39, opened 2023-11-28
Fix proof location enum

mfw78:
The proof location enum now matches that found in cow-sdk

The below description / changes can be used if more context is required...

# Description

The proof location enum is defined in the `cow-sdk` and not specified in the contract. Therefore this should match the `cow-sdk`.

# Changes

- [x] Documentation matches `cow-sdk`

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -3,6 +3,16 @@
 	"image": "mcr.microsoft.com/devcontainers/base:0",
 
 	"features": {
-		"ghcr.io/devcontainers/features/node:1": {}
+		"ghcr.io/devcontainers/features/node:1": {},
+		"ghcr.io/devcontainers/features/common-utils:2": {
+			"installZsh": true,
+			"configureZshAsDefaultShell": true,
+			"installOhMyZsh": true,
+			"installOhMyZshConfig": true,
+			"upgradePackages": true,
+			"username": "automatic",
+			"userUid": "automatic",
+			"userGid": "automatic"
+		  }
 	}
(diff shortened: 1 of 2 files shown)


fedgiac:
How can I verify this content? I checked the sdk and I confirmed [here](https://github.com/cowprotocol/cow-sdk/blob/7ef442e79add307237e617df4faf08a0f57fa2ad/src/composable/types.ts#L19-L32) that the numbering is right. But how can I check the data encoding for each variant? The most relevant piece of code appears to be [this](https://github.com/cowprotocol/cow-sdk/blob/7ef442e79add307237e617df4faf08a0f57fa2ad/src/composable/Multiplexer.ts#L280-L303), but it doesn't look like the uploader functions have been implemented yet. I also checked out the watchtower code but didn't find any reference to Swarm or Waku.

mfw78:
This would actually be the point at which it is formalised (ie. you're right, besides the `cow-sdk`, there is no point where the locations are specified / defined).

fedgiac:
If it's unused, should we specify it now? I suppose we'd like the watchtower to pick up proofs from there eventually, and maybe we'll wish to change the format. It also leads the reader to believe that they can upload the proofs to these services and they will be automagically picked up.

What about something like this:
```
| `SWARM` | `2` | Not yet implemented; reserved for future use. |
| `WAKU` | `3` | Not yet implemented; reserved for future use. |
```

fedgiac (approved):
Anyway, the changes are good and the encoding is tangential to the purpose of this PR, which should be merged as is.

PR #47, opened 2023-11-29
List CoW Swap features by priority

MartinquaXD:
Seems weird to have the IMO most insignificant feature listed first. Adjusted the order to based on my perceived usefulness for a trading venue.


mfw78 (approved):
:laughing:

PR #59, opened 2023-11-29
Bigger changes in concepts -> batch-auctions

MartinquaXD:
1 paragraph suggested that the number or frequency of batch auctions is influenced by the order duration of each user which is not the case

And the advantages bullet points seemed pretty repetitive and vague. My proposed wording is not great but roughly indicates what I think would be important benefits of batch auctions.

On an related note: That page has a drop down menu "On this page" with a single entry. Can this be removed until we have more items in the page?


mfw78 (approved):

mfw78:
FYI, the drop down navigation items etc, are all automatically generated, and these will be tidied up upon completion of the content :smiley:

PR #74, opened 2023-11-29
Rephrase token distribution

fedgiac:
Changes:
- Remove the repetition in the list. It read awkward to me as every line attempts to find a different word for "allocated to" from the treasury, which hinders readability (for example, `set aside` makes me thing that these tokens are reserved for future use, while they were indeed allocated) and has unnecessary redundancy.
- Put airdrops before investments. (See next point.)
- Explain what investments are and how they are assigned. It sounded very much not transparent, like if we were assigning them to our buddies in the community without any oversight.


fedgiac:

mfw78 (approved):

PR #91, opened 2023-11-29
expanded the description of the multiple ways in which we provide MEV…

acanidio-econ:
… protection

# Description

added the fact that solvers are responsible in case they get sandwiched - users are protected


acanidio-econ:

mfw78 (approved):

PR #94, opened 2023-11-29
Update uniform-clearing-prices.md

fleupold:
# Description

I wasn't happy with the phrasing of how UCP could be achieved on Ethereum in theory today. Frankly, I think it's incorrect. Transactions will always be executed sequentially against the most recent state and thus achieving it on the core protocol level is impossible. Even if a "close to optimal" ordering of txs existed, it's very unstable (every change to the ordering would break UCP)

# Changes

- [x] Rephrase how uniform clearing prices could be achieved in theory on Ethereum today
- [x] Highlight how the current incentives achieve the exact opposite
- [x] #99


fleupold:

mfw78 (approved):

mfw78:
Last item is tracked in #99 to roll across the repo

PR #101, opened 2023-11-29
Mention that the API does not accept exploitable orders

MartinquaXD:
Also Anxo would probably mad that you suggest putting random bytes in the `appdata` (or rather a hash that does not point to a document with the expected JSON format). 😄


mfw78:
FYI, I haven't edited the tutorials section (and I imagine is an area of active improvement the JS squad may want to work on), these are just copied straight from the old gitbook so they haven't been critically groomed yet. Part of other forthcoming work. `appData` and all :sweat_smile:

mfw78 (approved):

PR #140, opened 2023-11-29
Update app_data.mdx

fleupold:
# Description

Some suggestions for simplification (feel free to reject. Having the schema pulled in automatically is awesome by the way and gives me hope we can actually keep this doc up to date (not if we duplicate the entries in writing though)

# Changes

- [ ] Remove redundant information which is also part of Schema
- [ ] Clarify what is needed when not using the SDK or app-data utility (maybe we should make a tutorial for that as well)


mfw78:
Big fan of red PRs. Tutorial is tracked in cowprotocol/tutorial#12.

mfw78 (approved):

PR #141, opened 2023-11-29
Update signing_schemes.mdx

fleupold:
# Description

# Changes

- [x] Add a handly little contract I deployed for computing order UIDs and encodings using the Etherscan UI
- [x] Remove wrong caution that you cannot place a pre-sign order before it's confirmed on-chain (this is how the Safe UI currently does it)
- [x] Remove redundant graph that was not containing any additional information compared to the existing table

### DIFF
--- a/docs/cow-protocol/reference/core/signing_schemes.mdx
+++ b/docs/cow-protocol/reference/core/signing_schemes.mdx
@@ -23,7 +23,7 @@ When computing the domain separator, the following parameters are used:
   name: "Gnosis Protocol",
   version: "v2",
   chainId: /* chain ID for the current network: eg. 1 for mainnet */,
-  verifyingContract: /* address of the settlement contract */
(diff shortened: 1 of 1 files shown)


mfw78 (approved):

mfw78:
@fedgiac It may be handy if we also deploy this helper contract to sepolia as well?

fleupold:
> @fedgiac It may be handy if we also deploy this helper contract to sepolia as well?

We could maybe add it to the contracts repo and have it have a deterministic address as well.

fedgiac:
I added an issue to track this [here](https://github.com/cowprotocol/contracts/issues/67).
Where does it come from? Is it just a one-off untracked contract?

In terms of timing, once Sepolia is up and running I'll update the docs, so I'll see this again and deploy this contract.

fleupold:
Yes, I wrote it in the Remix IDE when debugging some smart order one day. Code is verified on etherscan though, so should be easy to add the the contracts repo

PR #145, opened 2023-11-29
Update README.mdx

fleupold:
# Description

ETH Flow has different deployments between prod and staging (so that orders placed on the staging UI get indexed by the staging autopilot).

I wonder how this is not an issue for TWAP (ie. how does the watchtower know into which API to place the order?)


mfw78:
The watch-tower assumes `prod` API for all orders. Also, I double-checked the user-interface, and it seems in TWAP within the CoW Swap `Safe` app, if one selects the native token, it immediately changes to the wrapped version of the token. Maybe @anxolin / @alfetopito can chime in in comments with respect to this.

mfw78 (approved):

anxolin:
Yes. I always defended as there's one contract, we should have only one environment. If we deploy a CoW Protocol v2, and if we want to keep `barn` environment, i would like to suggest we have a different deployment contract for it, so there's clear separation.

Watch Tower monitors the Composable CoW contract, as there's only a single instance, it post only to PROD. 
If we wanted staging here, i would suggest to re-deploy Composable CoW for Barn, however, i don't think is worth it. 

The UI knows TWAPs are in prod only, so even in Barn we use the PROD API for these orders.

PR #182, opened 2023-11-30
remove a constraint (block size) add a constraint (enough endowement)

acanidio-econ:
# Description
In the description of when a solution is valid, we used to say that it has to respect the block size. It seems redundant to specify that the solution needs to satisfy the basic constraints given by the blockchain (which, BTW, are more than simply block size). On the other hand, Felix L though we should specify that solvers also need to check that trader have enough tokens in their wallet

# Changes

- [x] remove a constraint (block size)
- [x] add a constraint (users' endowments)


mfw78 (approved):

mfw78:
Related issue #135

PR #194, opened 2023-12-01
fix: lazy-load version

mfw78:
# Description

Currently docusaurus `v2.4.3` is in use, despite `v3.0.0` being the `latest`. Lazy load images with `v3.0.0` was causing warning issues in the docusaurus compilation.

# Changes

- [x] Dropped `@docusaurus/plugin-ideal-image` to `v2.4.3`

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
   },
   "dependencies": {
     "@docusaurus/core": "2.4.3",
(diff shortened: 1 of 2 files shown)


mfw78:
Minor change, merging without review approval to avoid cascading PR conflicts

PR #243, opened 2023-12-18
chore: add chain-specific clarifications

### LINKED ISSUE #136: chore: add specification of which AMMs need to be at least considered in solution (EBBO)
https://github.com/cowprotocol/docs-v2/blob/ae8e1316e4db5ff899871d76dbb3824be82f506b/docs/cow-protocol/reference/core/auctions/social_consensus.md?plain=1#L20

Probably not having some random AMM on some exotic token qualifies here, instead we should define a list of protocols and base token pairs that need to be considered?

### LINKED ISSUE #231: chore: Should the core part of the solver competition specifications explicitly distinguish between general rules and rules that apply only on mainnet?
I just realized that the whole solver competition mechanism part  assumes we are talking about mainnet. Although i think it is fine for now, we do, for example, also run batch auctions on Gnosis Chain (and hope to expand in other chains as well). Do you think it's worth adding a few sentences here and there to clarify things and distinguish between different chains?

cc @fleupold @mfw78

harisang:
This PR adds a few minor clarifications here and there so that it is easier to know what applies on Mainnet vs Gnosis Chain.

Fixes #231
Fixes #136

### DIFF
--- a/docs/cow-protocol/reference/core/auctions/competition_rules.md
+++ b/docs/cow-protocol/reference/core/auctions/competition_rules.md
@@ -37,27 +37,27 @@ At CoW DAO's discretion, systematic violation of these rules may lead to penaliz
 
 :::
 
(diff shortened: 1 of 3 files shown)


docs/cow-protocol/reference/core/auctions/competition_rules.md
```diff
@@ -37,15 +37,34 @@ At CoW DAO's discretion, systematic violation of these rules may lead to penaliz
...
+  - 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2
```
fhenneke: Might be helpful to have the names of the tokens here as well.
(I did not check all the addresses.)

fhenneke (approved):
Changes look good to me.

docs/cow-protocol/reference/core/auctions/competition_rules.md
harisang: Addressed it.

PR #247, opened 2023-12-19
Add main sequence diagram for arbitrating auctions

sunce86:
# Description

Implements first part of the https://github.com/cowprotocol/docs-v2/issues/238:

> Could we add a diagram that shows the user, the orderbook, the autopilot and solvers (without differentiating between driver and engines)?


docs/cow-protocol/tutorials/arbitrate/README.mdx
```diff
@@ -53,6 +53,39 @@ They receive the pre-processed auction instance from the driver sub-component an
...
+    User->>orderbook: /post_order
+    autopilot->>orderbook: Request solvable orders
+    activate autopilot
+    activate orderbook
+    orderbook->>autopilot: Response solvable orders
+    deactivate orderbook
+    Note over autopilot: Create auction
+    autopilot->>solver: /solve
+    activate solver
+    solver-->>autopilot: Solution
```
MartinquaXD: We should probably use a different word than `solution` because the next call fetches the "Solution with calldata" which begs the question why we split it into 2 calls in the first place.
```suggestion
    solver-->>autopilot: Score
```

docs/cow-protocol/tutorials/arbitrate/README.mdx
```diff
@@ -53,6 +53,39 @@ They receive the pre-processed auction instance from the driver sub-component an
...
+    actor User
+    participant orderbook
+    participant autopilot
+    participant solver
+
+    User->>orderbook: /post_order
+    autopilot->>orderbook: Request solvable orders
+    activate autopilot
+    activate orderbook
+    orderbook->>autopilot: Response solvable orders
```
MartinquaXD:
```suggestion
    orderbook-->>autopilot: Response solvable orders
```

docs/cow-protocol/tutorials/arbitrate/README.mdx
```diff
@@ -53,6 +53,39 @@ They receive the pre-processed auction instance from the driver sub-component an
...
 The main flow of arbitrating an auction end-to-end can be seen in the following sequence diagram:
 
+```mermaid
+sequenceDiagram
+    actor User
+    participant orderbook
+    participant autopilot
+    participant solver
+
+    User->>orderbook: /post_order
```
MartinquaXD: All the other endpoints actually exist but not this one.
```suggestion
    User->>orderbook: POST /order
```

docs/cow-protocol/tutorials/arbitrate/README.mdx
MartinquaXD: Should we also show the DB here since the autopilot is actually just talking to that instead of the orderbook or would that go into too much detail?
Not sure what level of detail we want to have here.

docs/cow-protocol/tutorials/arbitrate/README.mdx
```diff
@@ -133,3 +133,125 @@ sequenceDiagram
     Note over autopilot: End
 ```
 
+## Run locally end to end
+
+In this tutorial we will run all the components locally on Ethereum mainnet, including the orderbook, autopilot, driver and solver engine. All these components have their own configuration parameters. For the sake of simplicity, we will use the bare minimum configuration for all of them.
+
+## Solver engine configuration file
+
+baseline.toml
```
MartinquaXD: Thief! 😄 
To make this easier to run we can also consider writing the config files inside `run.sh`. Then the user only has to copy and paste a single file.
Not sure what is nicer for a tutorial, though.

sunce86: gollum

sunce86: cat

docs/cow-protocol/tutorials/arbitrate/README.mdx
```diff
@@ -133,3 +133,125 @@ sequenceDiagram
...
+```
+chain-id = "1" # Ethereum mainnet
+base-tokens = [
+    "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2",
```
MartinquaXD: Would be nice to add the name of the tokens as a comment.

docs/cow-protocol/tutorials/arbitrate/README.mdx
```diff
@@ -133,3 +133,125 @@ sequenceDiagram
...
+# Fetch current auction from public orderbook and run auction locally
+# If you want to run an orderbook locally comment out this line
+# export SHADOW='https://api.cow.fi/mainnet';
+
+export SETTLE_INTERVAL=5
+export GAS_ESTIMATORS='Native,Web3';
+export PRICE_ESTIMATORS='None';
+export BLOCK_STREAM_POLL_INTERVAL_SECONDS='1'
+export TOKEN_OWNER_FINDERS='liquidity,blockscout,ethplorer,solvers';
+export SOLVER_TOKEN_OWNERS_CACHE_UPDATE_INTERVALS="300"
```
MartinquaXD: We can drop the token owners cache thing.
I only had it in my example because the yearn team needed some help setting this up and they relied on this feature.

docs/cow-protocol/tutorials/arbitrate/README.mdx
```diff
@@ -133,3 +133,125 @@ sequenceDiagram
...
+export SOLVER_TIME_LIMIT=15
+export PRICE_ESTIMATION_DRIVERS="baseline|http://localhost:11088/baseline"
+export DRIVERS="baseline|http://localhost:11088/baseline"
+
+# Orderbook args
+# Don't use built-in estimators; instead rely on the drivers
+export PRICE_ESTIMATORS='None'
+export NATIVE_PRICE_ESTIMATORS='baseline'
+# Run the bad token test to detect whether a token is supported
+export SKIP_TRACE_API='false'
```
MartinquaXD: I believe we actually no longer have this argument. 😅

docs/cow-protocol/tutorials/arbitrate/README.mdx
fleupold: Since the text is talking about three main components, I actually like that the diagram focusses on that. Will move the diagram up to the architecture section.

fleupold (commented):
I know I was asking for the "run" instructions, but I'm worried those will outdate faster than we can type whitespaces. @mfw78 do you have a suggestion on how we can refer to a place where developers can see an up to date instruction on how to run the services? Maybe the readme or the devcontainer file you started to work on?

fleupold:
We have it in version control, but I removed the detailed diagram as it doesn't maintain the driver/solver subdifferentiation and we have that information in the sections below. Happy to put it back if you feel it was better before.

For running locally it would be nice to link to instructions on how to run both the "shadow mode" as well as an oderbook + system (eventually also with the UI running)

docs/cow-protocol/tutorials/arbitrate/README.mdx
sunce86: Added.

docs/cow-protocol/tutorials/arbitrate/README.mdx
sunce86: Removed.

sunce86:
> I know I was asking for the "run" instructions, but I'm worried those will outdate faster than we can type whitespaces.

I believe this simplified example is probably most important among all tutorials and should be kept up to date (and probably be part of the services readme as you suggested). One thing we could do is to update the `services` PR template and add the acceptance tests part where the "check if docs need to be updated" would need to be verified before merging (although error prone but better than nothing).

mfw78:
> > I know I was asking for the "run" instructions, but I'm worried those will outdate faster than we can type whitespaces.
> 
> I believe this simplified example is probably most important among all tutorials and should be kept up to date (and probably be part of the services readme as you suggested). One thing we could do is to update the `services` PR template and add the acceptance tests part where the "check if docs need to be updated" would need to be verified before merging (although error prone but better than nothing).

I'm currently pulling this all into a `docker-compose` playground for integration within the `services` repo.

fleupold (approved):
LGTM, leaving it up to @mfw78 to weigh on if we want to keep the run instructions (which are quite detailed and maybe a link to instructions that will be more reasonably kept up to date may make more sense)

docs/cow-protocol/tutorials/arbitrate/README.mdx
```diff
@@ -49,54 +79,121 @@ If chosen as a winner of the auction, the driver is also responsible for getting
...
+export TRACING_NODE_URL=$ETHRPC;
+
+# Fetch current auction from public orderbook and run auction locally
+# If you want to run an orderbook locally comment out this line
+# export SHADOW='https://api.cow.fi/mainnet';
+
+export SETTLE_INTERVAL=5
+export GAS_ESTIMATORS='Native,Web3';
+export PRICE_ESTIMATORS='None';
+export BLOCK_STREAM_POLL_INTERVAL_SECONDS='1'
```
MartinquaXD: BTW the `_SECONDS` arguments will be outdated when https://github.com/cowprotocol/services/pull/2194 is merged.

MartinquaXD (approved):

mfw78 (approved):
I've removed the run script and driver config files and made a note on https://github.com/cowprotocol/services/pull/2228 that pointing to the playground readme needs to be documented here when it's pushed.

Issue #253, opened 2023-12-20
bug: wrong colors when you search for a page


anxolin:
### Problem

Review colors of the SEARCH functionality:

<img width="852" alt="image" src="https://github.com/cowprotocol/docs-v2/assets/2352112/37dc5664-698e-406a-a9cf-9e36f3a4bd76">

Hard to read text

Issue #241, opened 2023-12-18
bug: no top/bottom padding after an <Image>


alfetopito:
### Problem

Seems like <Image> components have no padding on top/bottom

### To reproduce

If you can reproduce the behaviour, steps to reproduce:

1. Add and `<Image>` element
2. After it, add a `:::tip` block
3. Visit the page where the image is followed by the tip block

### Expected behaviour

There should be some padding after the image

### Screenshots/logs

Using the `<Image>` tag, followed by a `:::tip`

![image](https://github.com/cowprotocol/docs-v2/assets/43217/66a2aac3-99f5-4d75-8251-901c105da2d7)

The same, replacing `<Image>` with native `![alt](/localtion)` markdown notation

![image](https://github.com/cowprotocol/docs-v2/assets/43217/27cd2d52-d5c7-4fbd-9d95-592b9b44171b)

### docs version/commit hash

First noticed on this commit 16b485fc334da64cd246f783753033563710d8f1

### Additional context

Discussion initiated [in this thread](https://cowservices.slack.com/archives/C04N05LJK54/p1702922460712259)

anxolin:
Ideally we add the label so it shows in the kanban when we filter by our team

Issue #45, opened 2023-11-29
chore: Make links stand out visually


MartinquaXD:
### Background

The docs are super nice and use links to get you to places with more info on some topics but because links are only identifiable as such by hovering over them you can probably not notice lots of links.
Basically you currently have to hover over every word in order to not miss any links.

### Acceptance criteria

Links are immediately identifiable as such (or at least you should not have to hover over them to detect them).

fairlighteth:
Is this issue still relevant? This is how links currently are rendered:

---

![Image](https://github.com/cowprotocol/docs-v2/assets/31534717/f39a175d-6f12-4212-bd8c-49fb5a0ffb80)

![Image](https://github.com/cowprotocol/docs-v2/assets/31534717/0049c587-c0c1-42ea-8ee0-d1725e9c6e95)

PR #258, opened 2023-12-28
feat: styling fixes

fairlighteth:
# Description

Fixes:
- #253 
- #241 

Also #45 seems not an actual issue anymore? Therefore not addressed here.

### DIFF
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -52,6 +52,7 @@ html[data-theme='dark']:root {
   --ifm-background-color: #061224;
   --ifm-background-surface-color: #081d39;
   --ifm-color-content: var(--ifm-color-primary);
+  --docsearch-highlight-color: var(--ifm-color-primary-lighter);
   
 }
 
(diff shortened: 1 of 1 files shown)


mfw78:
I notice some weird effects with margins on the images:

1. Load a page which contains some lazy loaded images
2. On the condition whereby the lazy loaded image is not loaded (blurred placeholder): click the image to load it
3. Observe that when the image loads, it subsequently modifies the bottom margin to the image

anxolin (approved):
Soft approve, as there's a pending comment from mfw

@fairlighteth fixes the issue in the search box!

For the margin. I think @mfw78 is right. Its very subtle. I reproduced it by loading any page, like https://docs-git-styling-fixes-cowswap.vercel.app/cow-protocol/tutorials/cow-swap/native , and without scrolling, just disable your internet. Then scroll to a image that hasn't been loaded, enable internet, and click on the image.

fairlighteth:
IMO the specific scenario @mfw78 mentions should be addressed separately (if at all..). Main reason is that the blurry image seems to be a placeholder SVG. We could target it with margin, but consider it out of scope here.

Screenshot 2024-01-08 at 13 26 27

PR #288, opened 2024-01-18
Update mission.md

marshymarsh:
Update to be less protocol, more DAO

# Description

# Changes

- [ ] ...
- [ ] ...


mfw78 (approved):

mfw78:
In general, I think we should always mention "CoW Protocol" prior to "CoW Swap" (as the latter cannot exist without the former).

PR #292, opened 2024-01-23
feat: fees

marshymarsh:
# Description

new fees section

# Changes

added fees page to governance section


mfw78 (approved):
Looks good. Made a minor change to the note so as to use admonitions per site-wide usage.

Issue #139, opened 2023-11-29
feat: use new solvers schema for driver <> solver_engine interface


fleupold:
### Problem

The legacy `instance.json` format is terribly redundant, full of inconsistencies and unnecessary fields and was refactored during the solver-driver colocation project in favour of a new/cleaner interface between driver and solver_engines (https://github.com/cowprotocol/services/blob/a72c4c374afea785929a0a0a98d7840faf3df34c/crates/solvers/openapi.yml)

### Suggested solution

Explain the problem/result file in terms of the new data model and mark the existing document as deprecated (or remove altogether)

### Alternatives considered

Keep maintaining the old json (but I think that's a bad idea)

mfw78:
In regards to scoping for this, would this be a joint effort between backend and solvers?

fleupold:
I think it should be driven by the solver team with the backend team providing clarification and help where needed. The interface itself is documented using swagger (and in code).
Maybe @fhenneke's test solver from dappcon could be adopted to work with the new interface.

fhenneke:
One thing blocking me from changing the documentation is that the code and api specification for the new interface are inconsistent, see [this issue](https://github.com/cowprotocol/services/issues/2182).
I suppose the rust code is the reference then. But I am not familiar enough with it to derive the format of request and response.

PR #324, opened 2024-02-09
[Solver team] Add new solver api info

harisang:
This PR updates the documentation regarding the solver/driver api, and addresses issue #139 .

### DIFF
--- a/docs/cow-protocol/reference/core/auctions/schema.md
+++ b/docs/cow-protocol/reference/core/auctions/schema.md
@@ -62,37 +62,13 @@ We now share two example token entries corresponding to [WETH](https://etherscan
     "trusted": true
 }
 ```
-We clarify a few things regarding the entries above.
-
-- The referencePrice of a token, viewed in isolation, is just a number and does not have inherent meaning. However, the referencePrice suggests an (approximate) exchange rate between two tokens, and this is the reason these prices are used to convert everything to WETH. Specifically, we usually have that `WETH` has a referencePrice of 1000000000000000000, which should be interpreted as 1 wei = 1/10<sup>18</sup> `WETH` (since `WETH` has 18 `decimals`) has a price of 1000000000000000000.
-- The `referencePrice` of `USDC` is set to 449666048539228625975640064, which means that one atom of USDC, i.e., 1/10<sup>6</sup> of 1 `USDC` has a price of 449666048539228625975640064, given that 1 wei has price 1000000000000000000. This suggests that 1 USDC is equal to referencePrice(USDC) * 10<sup>6</sup> / (referencePrice(WETH) * 10<sup>18</sup>) = 0.0004496660485392286 WETH, or equivalently, that 1 WETH is equal to 2223.87259 USDC.
+We clarify a few things regarding the entries above. The referencePrice of a token, viewed in isolation, is just a number and does not have inherent meaning. However, the referencePrice suggests an (approximate) exchange rate between two tokens, and this is the reason these prices are used to convert a quantity denominated in some token to WETH. Specifically, we usually have that `WETH` has a referencePrice of 1000000000000000000, which should be interpreted as 1 wei = 1/10<sup>18</sup> `WETH` (since `WETH` has 18 `decimals`) has a price of 1000000000000000000. This implies that the `referencePrice` of `USDC`, which here is set to 449666048539228625975640064, means that one atom of USDC, i.e., 1/10<sup>6</sup> of 1 `USDC` has a price of 449666048539228625975640064, given that 1 wei has price 1000000000000000000. This suggests that 1 USDC is equal to referencePrice(USDC) * 10<sup>6</sup> / (referencePrice(WETH) * 10<sup>18</sup>) = 0.0004496660485392286 WETH, or equivalently, that 1 WETH is equal to 2223.87259 USDC.
 
 :::note
 
 Both tokens above are accepted for internalization.
 
 :::
-
-### `prices`
-
-This key specifies the uniform clearing price vector (UCP) that governs the proposed solution. It is a dictionary mapping each token traded by a user order to its computed price. Each price is an stringified unsigned integer, and the whole vector is scale invariant, meaning that multiplying each entry by the same constant will not change anything; thus the scaling is chosen by the solver. We also stress that a solution need only contain prices for the tokens appearing in the executed user orders, and that solvers are free to choose the unit of measure they will use.
-
-An example containing the computed prices of [USDC](https://etherscan.io/token/0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48), [COW](https://etherscan.io/token/0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB) and [WETH](https://etherscan.io/token/0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2) is given below.
-
-```json
-"prices": {
-    "0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48": "389193245742363509576247472",
-    "0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB": "137298311435590",
-    "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2": "1000000000000000000"
-}
-```
-
-The above entries should be interpreted as follows:
-
-- 1 wei (3rd entry) has a price of 1000000000000000000 = 10<sup>18</sup>.
-- The lowest denomination of `USDC` (1st entry), i.e., 1 / 10<sup>6</sup> of `USDC`, has a price of 389193245742363509576247472 relative to the price of 10<sup>18</sup> that wei has. This translates, for example, to 1 `WETH` having a price of $$\frac{1000000000000000000 \cdot 10^{18}}{389193245742363509576247472 \cdot 10^{6}} \approx 2569.417$$ `USDC`.
-- The lowest denomination of `COW` (2nd entry) is 1 / 10<sup>18</sup> , and it has a price of 137298311435590. This translates, for example, to 1 `COW` having a price of $$\frac{137298311435590 \cdot 10^{18}}{389193245742363509576247472 \cdot 10^{6}} \approx 0.35277$$ USDC.
-
 
 ### `orders`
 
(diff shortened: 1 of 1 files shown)


harisang:
@fhenneke Could you have a quick look up till the point I have marked to check for obvious issues/discrepancies?

docs/cow-protocol/reference/core/auctions/schema.md
```diff
@@ -1,12 +1,15 @@
 ---
 sidebar_position: 5
-draft: true
 ---
 
 # Schemas
 
 All the instances and solutions of the batch auction problem are formatted in [JSON](https://www.json.org/json-en.html). In this section, we describe these schemas.
 
+:::
```
fhenneke:
```suggestion
:::note
```
this prints strangely atm. did you want this to be a note?

docs/cow-protocol/reference/core/auctions/schema.md
```diff
@@ -15,441 +18,188 @@ To avoid precision loss, some numerical literals are encoded as strings, referre
...
-A Stable pool corresponds to [Curve/Balancer stable pools](https://dev.balancer.fi/resources/pool-interfacing/stable-pool), and has the following differences to the Constant Product pool described above,
+More information about the exact descriptions of these pools can be found [here](https://docs.cow.fi/cow-protocol/reference/apis/solver).
 
-- `"kind"`: the type is set to "Stable".
-- `"amplification_parameter"`: See [amplification coefficient](https://curve.fi/files/stableswap-paper.pdf).
-- `"scaling_rates"`: See [getScalingFactors](https://github.com/balancer-labs/balancer-v2-monorepo/blob/80b0e1b129d575c313f59800ec7e19237a43c087/pkg/pool-utils/contracts/BasePool.sol#L523-L525).
+### `effectiveGasPrice`
 
-An example of such an AMM between [DAI](https://etherscan.io/token/0x6b175474e89094c44da98b954eedeac495271d0f), [USDC](https://etherscan.io/token/0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48) and [USDT](https://etherscan.io/token/0xdac17f958d2ee523a2206206994597c13d831ec7) is given below.
+This key is a single entry that is a stringified integer describing the current gas price in wei.
```
fhenneke:
```suggestion
This key is a single entry that is a stringified integer describing the current gas price in wei.

:::note
The actual effective gas price when settling a solution might be different due to fluctuations of gas prices and due to the choice of submission strategy.
:::
```
we could add a comment here to emphazise that the gas price for submission of the settlement might be different even withoug changes of the gas price.

docs/cow-protocol/reference/core/auctions/schema.md
```diff
@@ -15,441 +18,188 @@ To avoid precision loss, some numerical literals are encoded as strings, referre
...
-- `"token"`: the address of the token that we authorize.
-- `"amount"`: a stringified integer that corresponds to the (minimum) amount we authorize the target contract to use. **We stress here** that currently the behavior is the following. If there is already sufficiently large (i.e., larger than the `"amount"` entry) allowance set by a previous settlement, then the approval is ignored altogether. On the other hand, if the current allowance is lower than `"amount"`, then the driver will set a maximum value for the approval (much larger than the actual `"amount"` specified by the entry); this allows for future gas savings.
+- `kind`: this is string of the set {"fulfillment", "jit"}, which corresponds to an order existing in the orderbook, or a just-in-time liquidity order placed by the solver, respectively.
+- `order`: in case of a "fulillment" trade, the `uid` of the order is provided here as a string. In case it is a just-in-time liquidity order, the specifications of the order are explicitly given as a dictionary; more details can be found [here](https://docs.cow.fi/cow-protocol/reference/apis/solver).
```
fhenneke:
```suggestion
- `order`: in case of a "fulfillment" trade, the `uid` of the order is provided here as a string. In case it is a just-in-time liquidity order, the specifications of the order are explicitly given as a dictionary; more details can be found [here](https://docs.cow.fi/cow-protocol/reference/apis/solver).
```

docs/cow-protocol/reference/core/auctions/schema.md
```diff
@@ -15,441 +18,188 @@ To avoid precision loss, some numerical literals are encoded as strings, referre
...
+- `fee`: this entry exists only for "fullilment" trades, and maps to a stringified integer describing the fee of the order (either pre-signed or solver computed), denominated in the sell token.
+- `executedAmount`: this is a stringified integer corresponding to the sell amount (for sell orders) or the buy amount (for buy orders) that would get executed. Note that this is amount is "separate" from the "fee" amount that was described above; this, for example, means that for a sell "fulfillment" order, the trader will send a total of `fee + executedAmount` sell tokens to the contract. We also stress that ii is the executedAmount where uniform clearing prices are being applied to.
```
fhenneke:
```suggestion
- `executedAmount`: this is a stringified integer corresponding to the sell amount (for sell orders) or the buy amount (for buy orders) that would get executed. Note that this is amount is "separate" from the "fee" amount that was described above; this, for example, means that for a sell "fulfillment" order, the trader will send a total of `fee + executedAmount` sell tokens to the contract. We also stress that it is the `executedAmount` where uniform clearing prices are being applied to.
```

docs/cow-protocol/reference/core/auctions/schema.md
```diff
@@ -15,441 +18,188 @@ To avoid precision loss, some numerical literals are encoded as strings, referre
...
+The score is a key that describes the "bid" a solver makes for the batch in the [solver auction](rewards), as it will get ranked according to it. The protocol picks the solution with the highest score, given that it is strictly positive. The score maps to a dictionary with two entries. The first entry is the following:
+
+- `kind`: this is a string of the set {"solver", "riskAdjusted"}, that determines whether a solver will explicitly provide a score or will delegate the score computation to the default driver by only specifying a probability of success for the proposed solution.
+
+If we have `"kind": "solver"`, then there is a second entry, labeled `score`, that corresponds to a stringified float that specifies the score attached to the solution. On the other hand, if we have `"kind": "riskAdjusted"`, then there is a second entry, labeled `successProbability`, that is a a stringified float that specifies the success probability of the proposed solution.
```
fhenneke:
```suggestion
If we have `"kind": "riskAdjusted"`, then there is a second entry, labeled `successProbability`, that is a a stringified float that specifies the success probability of the proposed solution. On the other hand, if we have `"kind": "solver"`, then there is a second entry, labeled `score`, that corresponds to a stringified float that specifies the score attached to the solution.

:::note
All solvers are encouraged to submit a score via `"kind": "riskAdjusted"` and `successProbability`. Submitting scores directly via `"kind": "solver"` is discouraged and this option might get deprecated.
:::
```
Not sure how to formulate this. I think we should explicitly discourage the submission of scores.

fhenneke (approved):
I left a few comment.
In general this looks good to me.

docs/cow-protocol/reference/core/auctions/schema.md
harisang: I actually changed the phrasing a bit. Let me know if you agree with the new version

docs/cow-protocol/reference/core/auctions/schema.md
harisang: Added some clarifications. Let me know if you think it's fine now.

docs/cow-protocol/reference/core/auctions/schema.md
```diff
@@ -15,441 +18,192 @@ To avoid precision loss, some numerical literals are encoded as strings, referre
...
-A Stable pool corresponds to [Curve/Balancer stable pools](https://dev.balancer.fi/resources/pool-interfacing/stable-pool), and has the following differences to the Constant Product pool described above,
+More information about the exact descriptions of these pools can be found [here](https://docs.cow.fi/cow-protocol/reference/apis/solver).
 
-- `"kind"`: the type is set to "Stable".
-- `"amplification_parameter"`: See [amplification coefficient](https://curve.fi/files/stableswap-paper.pdf).
-- `"scaling_rates"`: See [getScalingFactors](https://github.com/balancer-labs/balancer-v2-monorepo/blob/80b0e1b129d575c313f59800ec7e19237a43c087/pkg/pool-utils/contracts/BasePool.sol#L523-L525).
+### `effectiveGasPrice`
 
-An example of such an AMM between [DAI](https://etherscan.io/token/0x6b175474e89094c44da98b954eedeac495271d0f), [USDC](https://etherscan.io/token/0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48) and [USDT](https://etherscan.io/token/0xdac17f958d2ee523a2206206994597c13d831ec7) is given below.
+This key is a single entry that is a stringified integer describing an estimate of the effective gas price for the winning solution, denominated in wei.
```
fhenneke, line 122: Maybe it was better before. As in
```suggestion
This key is a single entry that is a stringified integer describing the current gas price in wei.
```

What I wanted to get at was that solvers interpreted the number as an estimate of the gas_price and that only general variations in gas price would lead to a different value for the settlement. But this is wrong, if I understand correctly. We do in some cases add 10% or something like that. This additional tip seems to be the main source of deviation and tripped up solvers in the past.

docs/cow-protocol/reference/core/auctions/schema.md
```diff
@@ -15,441 +18,192 @@ To avoid precision loss, some numerical literals are encoded as strings, referre
...
+The score is a key that describes the "bid" a solver makes for the batch in the [solver auction](rewards), as it will get ranked according to it. The protocol picks the solution with the highest score, given that it is strictly positive. The score maps to a dictionary with two entries. The first entry is the following:
+
+- `kind`: this is a string of the set {"solver", "riskAdjusted"}, that determines whether a solver will explicitly provide a score or will delegate the score computation to the default driver by only specifying a probability of success for the proposed solution.
+
+If we have `"kind": "solver"`, then there is a second entry, labeled `score`, that corresponds to a stringified float that specifies the score attached to the solution. On the other hand, if we have `"kind": "riskAdjusted"`, then there is a second entry, labeled `successProbability`, that is a a stringified float that specifies the success probability of the proposed solution.
```
fhenneke:
```suggestion
If we have `"kind": "solver"`, then there is a second entry, labeled `score`, that corresponds to a stringified float that specifies the score attached to the solution. On the other hand, if we have `"kind": "riskAdjusted"`, then there is a second entry, labeled `successProbability`, that is a stringified float that specifies the success probability of the proposed solution.
```

docs/cow-protocol/reference/core/auctions/schema.md
harisang: But i believe this is not just the base fee but it tries to approximate the effective gas price including the fee. Maybe I am wrong here. Let me try to check the code and get back to you.

harisang: https://github.com/cowprotocol/services/blob/06e9c4ed58da488973299c14164c452c59505d72/crates/driver/src/domain/eth/gas.rs#L66

This seems to suggest tip is already added here.

fhenneke: At some point is was base fee + "small tip", where the small tip was for inclusion of fully internalized transactions. This is then a good estimate for the case where there are no interactions, just transfers to users.
When there are interactions, we added an additional tip, so that block builders include our transaction faster or, rather, we pay them to not extract MEV. Not sure if that changed at some point. @sunce86 

Also not super important. I just remembered that solvers did have problems understanding the meaning which lead to "unintentional" overbidding.

harisang:
Merging as it is pressing to have some documentation around the new api, and we can revisit in case there is some inaccuracy

PR #326, opened 2024-02-16
Update definitions.md

olgafetisova:
# Description

Adding the following definitions:
- Quote Deviation

### DIFF
--- a/docs/cow-protocol/reference/core/definitions.md
+++ b/docs/cow-protocol/reference/core/definitions.md
@@ -20,19 +20,13 @@ A single order submitted to the CoW Protocol API (i.e. `GPv2Order.Data`)
 
 An intent by a smart-contract to place `0..n` discrete orders under some programmatic logic. For example, a conditional order may only be valid if the price of an asset reaches a certain level.
 
-#### Limit Order
-An order to swap specified `sellAmount` of `sellToken` for a specified `buyAmount` of `buyToken` or better.
-
-#### TWAP Order
-Time-Weighted Average Price order. It is an algorithmic trade execution strategy that aims to achieve an average execution price close to the time-weighted average price of the specified period.
-
 #### Surplus
 
 The price improvement on a user's limit price.
 
 #### Quote deviation
 
-The price difference between the quote and the executed price. May be used interchangeably with slippage.
+The difference between the quote the user received and the price at which their discrete order was settled.
 
 #### Gasless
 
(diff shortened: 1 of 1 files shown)


olgafetisova:

docs/cow-protocol/reference/core/definitions.md
```diff
@@ -20,10 +20,20 @@ A single order submitted to the CoW Protocol API (i.e. `GPv2Order.Data`)
 
 An intent by a smart-contract to place `0..n` discrete orders under some programmatic logic. For example, a conditional order may only be valid if the price of an asset reaches a certain level.
 
+#### Limit Order
+An order to swap specified `sellAmount` of `sellToken` for a specified `buyAmount` of `buyToken` or better.
+
+#### TWAP Order
+Time-Weighted Average Price order. It is an algorithmic trade execution strategy that aims to achieve an average execution price close to the time-weighted average price of the specified period.
```
mfw78:
```suggestion
```

In general, I don't think we should define "external" definitions (ie. non-specific to CoW Protocol / batch auctions). The unique thing about CoW is it gives surplus over limit orders, which is defined below in _surplus_. "Limit order" and "TWAP order" are known terms (or able to be googled). We should however IMO on the "Limit order" and/or "TWAP order" page within the concepts section link to a wikipedia article that is able to source the external definition.

docs/cow-protocol/reference/core/definitions.md
```diff
@@ -20,10 +20,20 @@ A single order submitted to the CoW Protocol API (i.e. `GPv2Order.Data`)
...
+
 #### Surplus
 
 The price improvement on a user's limit price.
 
+#### Quote deviation
+
+The price difference between the quote and the executed price. May be used interchangeably with slippage.
```
mfw78:
```suggestion
The difference between the quote the user received and the price at which their discrete order was settled. May be used interchangeably with slippage.
```

Just a minor adjustment to use definitions we've previously defined (ie. discrete order). Also, I'm not sure if this can be used interchangeably with slippage? I thought slippage represented the difference between the limit price and the executed price (ie. what is normally defined as "surplus")?

mfw78 (approved):

gogonimago:
Just one edit from me here, for the term "Gasless"

“An expression that does not require a user make a transaction on the blockchain, and therefore does not require the user to pay gas in the native token.” Should be “An expression that does not require a user to make a transaction on the blockchain, and therefore does not require the user to pay gas in the native token.” (Missing a “to” before “make”)

@mfw78

PR #331, opened 2024-02-23
Update cow-amm.md

acanidio-econ:
# Description

various edits

# Changes

- [x] the list of supported oracles had only Uniswap. But then in the description there is also Balancer. I added Balancer to the initial list
- [x] edit the very last sentence (the problem with contributing only one token to the AMM). It is not so much that the AMM will trade at an unfavorable price: the price is always the one coming out of the solvers' competition and should be correct. The problem is that it will immediately sell a fraction of what it just received, which may not be the intended behavior
- [x]  fixed some typos


harisang (approved):

harisang:
@acanidio-econ Looks good, and just checked that it is all good to be merged (no issues with deployment)

mfw78 (approved):

PR #347, opened 2024-04-03
docs: update widget docs

shoom3301:
# Description

1. `cowSwapWidget()` -> `createCowSwapWidget()`
2. `params` -> `{ params }`


anxolin (approved):
Not related to this PR, but, shouldn't we advertise the use of react-widget at the being of the doc?

If someone uses react they are way better off using the other library

anxolin:
Also, and also unrelated, but the configuration makes more sense at the end instead of in the middle, nup? 

image

shoom3301:
@anxolin thank you!
Both points make sense, will work on it in the next iteration

PR #348, opened 2024-04-03
Update partner-fee.md

gogonimago:
# Description

# Changes

- [ ] ...
- [ ] ...

### DIFF
--- a/docs/governance/fees/partner-fee.md
+++ b/docs/governance/fees/partner-fee.md
@@ -5,15 +5,14 @@ id: partner-fee
 
 ### Partner Fee Calculation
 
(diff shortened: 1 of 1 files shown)


mfw78 (approved):

anxolin (commented):
👍 Thanks!

Do you think this is easy to understand this article? 

For example, `The Partner Fee is calculated as the sum of all transactions traded through your integration with CoW Protocol with Partner` which is the opening phrase, seems not true

Also, the examples provided. They were extracted from some example we had in some internal documentation, but probably is not easy enough to understand?

gogonimago:
Yeah I was confused on that opening line... I'm not sure what exactly it means, especially with mentioning "Partner Fee" twice in seemingly different contexts. 

Did you write that or did someone else did

PR #373, opened 2024-04-22
Update sending-transactions-via-rpc.md

acanidio-econ:
refer to the correct page where we discuss the endpoints

# Description

# Changes

- [ ] ...
- [ ] ...

### DIFF
--- a/docs/mevblocker/users-and-integrators/users/sending-transactions-via-rpc.md
+++ b/docs/mevblocker/users-and-integrators/users/sending-transactions-via-rpc.md
@@ -4,7 +4,7 @@ sidebar_position: 3
 
 # Sending Transactions via RPC
 
-MEV Blocker supports sending transactions via the standard `eth_sendRawTransaction` method.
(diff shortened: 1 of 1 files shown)


docs/mevblocker/users-and-integrators/users/sending-transactions-via-rpc.md
```diff
@@ -4,7 +4,7 @@ sidebar_position: 3
 
 # Sending Transactions via RPC
 
-MEV Blocker supports sending transactions via the standard `eth_sendRawTransaction` method.
+MEV Blocker supports sending transactions via multiple endpoints (as discussed [here](https://docs.cow.fi/mevblocker/concepts/endpoints)) using the standard `eth_sendRawTransaction` method.
```
mfw78:
```suggestion
MEV Blocker supports sending transactions via multiple endpoints (as discussed [here](/mevblocker/concepts/endpoints)) using the standard `eth_sendRawTransaction` method.
```

Please don't use absolute links as this will break things if we need to move the documentation site in the future :pray:

mfw78 (changes requested):

mfw78 (approved):

PR #392, opened 2024-06-13
docs(widget): flexible config for partner fee params

shoom3301:
# Description

Since partner fee can be set for any trade type on any network, we should provide an option to flexibly configurate the fee.

Now both `bps` and `recipient` have type `FlexibleConfig`.

```typescript
export type PerTradeTypeConfig<T> = Record<TradeType, T>

export type PerNetworkConfig<T> = Record<SupportedChainId, T>

export type FlexibleConfig<T> =
  | T
  | PerNetworkConfig<T>
  | PerTradeTypeConfig<T>
  | PerTradeTypeConfig<PerNetworkConfig<T>>
  | PerNetworkConfig<PerTradeTypeConfig<T>>
```

### DIFF
--- a/docs/cow-protocol/tutorials/widget/widget.md
+++ b/docs/cow-protocol/tutorials/widget/widget.md
@@ -69,6 +69,62 @@ const params: CowSwapWidgetParams = {
 createCowSwapWidget(widgetContainer, { params })
 ```
 
+### Flexible config
+
+Both `bps` and `recipient` can be set for different chains and different trade types (swap/limit/advanced). 
+
+Bellow you can see the `partnerFee` configuration variations:
+
+```typescript
+import {PartnerFee, SupportedChainId, TradeType} from '@cowprotocol/widget-lib'
+
+// The fee is 1% for all trades on all chains
+const a: PartnerFee = {
+    bps: 100,
+    recipient: '0x0000000000000000000000000000000000000000'
+}
+
+// Per network and per trade type
+const b: PartnerFee = {
+    bps: {
+        [SupportedChainId.MAINNET]: {
+            [TradeType.SWAP]: 100, [TradeType.LIMIT]: 50, [TradeType.ADVANCED]: 30,
+        },
+        [SupportedChainId.ARBITRUM_ONE]: {
+            [TradeType.SWAP]: 100, [TradeType.LIMIT]: 50, [TradeType.ADVANCED]: 30,
+        },
+        [SupportedChainId.GNOSIS_CHAIN]: {
+            [TradeType.SWAP]: 100, [TradeType.LIMIT]: 50, [TradeType.ADVANCED]: 30,
+        },
+        [SupportedChainId.SEPOLIA]: {
+            [TradeType.SWAP]: 100, [TradeType.LIMIT]: 50, [TradeType.ADVANCED]: 30,
+        },
+    },
+    recipient: '0x0000000000000000000000000000000000000000',
+}
+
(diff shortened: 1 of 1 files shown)


docs/cow-protocol/tutorials/widget/widget.md
```diff
@@ -69,6 +69,62 @@ const params: CowSwapWidgetParams = {
...
+        },
+        [TradeType.ADVANCED]: {
+            [SupportedChainId.MAINNET]: '0x...j', [SupportedChainId.ARBITRUM_ONE]: '0x...i',
+            [SupportedChainId.GNOSIS_CHAIN]: '0x...k', [SupportedChainId.SEPOLIA]: '0x...l',
+        },
+    }
+}
+```
+
+See [FlexibleConfig](https://github.com/cowprotocol/cowswap/blob/develop/libs/widget-lib/src/types.ts) type for more information.
```
anxolin, line 126: I think we should add a note on how to do even more flexible ones using the listeners. I believe this config will cover 90% of integrations, but is still nice to point out that we have even a more fine level of granularity allowing you to set configurations based on flow


This actually makes me think if we should allow to provide a configuration at some point using a function just to make it even easier.

How it works is, the user defines a function that receive all relevant data: network, trade params, flow, etc.
They return the partnerFee configuration

Our widget library would remember the last partnerFee reported to the widget, and if it changed, it would send the new one. This way, we don't notify changes all the time, but let the user have a very flexible logic.

anxolin (commented):
Looks great, mostly a nitpick

docs/cow-protocol/tutorials/widget/widget.md
```diff
@@ -69,6 +69,62 @@ const params: CowSwapWidgetParams = {
...
+```typescript
+import {PartnerFee, SupportedChainId, TradeType} from '@cowprotocol/widget-lib'
+
+// The fee is 1% for all trades on all chains
+const a: PartnerFee = {
+    bps: 100,
+    recipient: '0x0000000000000000000000000000000000000000'
+}
+
+// Per network and per trade type
```
alfetopito: The comment is the same as the one below, but both are different

```suggestion
// Different fee per network and per trade type, same recipient for all
```

docs/cow-protocol/tutorials/widget/widget.md
```diff
@@ -69,6 +69,62 @@ const params: CowSwapWidgetParams = {
...
+            [TradeType.SWAP]: 100, [TradeType.LIMIT]: 50, [TradeType.ADVANCED]: 30,
+        },
+        [SupportedChainId.SEPOLIA]: {
+            [TradeType.SWAP]: 100, [TradeType.LIMIT]: 50, [TradeType.ADVANCED]: 30,
+        },
+    },
+    recipient: '0x0000000000000000000000000000000000000000',
+}
+
+// Per trade type and per network
```
alfetopito:
```suggestion
// Same fee for all, different recipient per network and per trade type
```

alfetopito (approved):

docs/cow-protocol/tutorials/widget/widget.md
shoom3301: https://github.com/cowprotocol/docs/pull/392/commits/744105cde8a3e3469b99f26ebcdec9284a3dd7ac

anxolin (approved):

cowmarketing (approved):

PR #404, opened 2024-07-25
Update subscription-fees.md

acanidio-econ:
I corrected the fact that we also charge for flow that is not exclusive because it is sent via different private RPC points. The math is correct but the description in the initial paragraph was not

# Description

# Changes

- [ ] ...
- [ ] ...

### DIFF
--- a/docs/mevblocker/builders/fees/subscription-fees.md
+++ b/docs/mevblocker/builders/fees/subscription-fees.md
@@ -4,7 +4,7 @@ sidebar_position: 1
 
 # The Fee
 
-All builders connected to MEV Blocker will pay a per-block-won fee recomputed every period (for example, every month). In practice, at the beginning of period $t$ a builder decides whether to receive flow from MEV Blocker. If it decides to do so, for every block won during period $t$ it will pay a per-block fee calculated as a percentage of the the average per-block-won value of exclusive MEV Blocker transactions during the previous period. Mathematically, the per-block fee for period $t$ is a percentage of:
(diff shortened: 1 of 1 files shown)


cowmarketing (approved):

docs/mevblocker/builders/fees/subscription-fees.md
```diff
@@ -4,7 +4,7 @@ sidebar_position: 1
 
 # The Fee
 
-All builders connected to MEV Blocker will pay a per-block-won fee recomputed every period (for example, every month). In practice, at the beginning of period $t$ a builder decides whether to receive flow from MEV Blocker. If it decides to do so, for every block won during period $t$ it will pay a per-block fee calculated as a percentage of the the average per-block-won value of exclusive MEV Blocker transactions during the previous period. Mathematically, the per-block fee for period $t$ is a percentage of:
+All builders connected to MEV Blocker will pay a per-block-won fee recomputed every period (for example, every month). In practice, at the beginning of period $t$ a builder decides whether to receive flow from MEV Blocker. If it decides to do so, for every block won during period $t$, it will pay a per-block fee calculated as a percentage of the average per-block-won value of MEV Blocker transactions during the previous period. Mathematically, the per-block fee for period $t$ is a percentage of:
```
cmagan, line 7: I think the fix is important.
I wonder if we have any process for introducing such changes?
Should we notify affected parties like builders, or should the public github PR process should be enough?

cmagan (approved):

PR #447, opened 2025-01-16
Update partner-fee.md

marshymarsh:
15% to 25% as default

# Description

# Changes

- [ ] ...
- [ ] ...


harisang:
Is this changing, and if so, when? We would need to do some adjustments in the accounting side, this is why i am asking

cowmarketing:
No longer aplicable, overiden by PR #460

PR #458, opened 2025-03-17
Update ETH-flow comment

fedgiac:
# Description

I'd like to remove from the ETH-flow entry the description of it being "deterministic."
Most other contracts in this page are "deterministic" in the loose meaning of this word deriving from the banner above, but we don't bother mentioning it.
The ETH-flow is the one that is most controversial, because "deterministic" is normally implied to mean "deployed with `CREATE2`," and that's not the case for the ETH flow (see [deployment](https://github.com/cowprotocol/ethflowcontract/commit/e1817b312132dd0d9e8952deae9fc71549558f09)).
Reading the table entry without reading the banner first would be misleading (and I actually was mislead by that, I started this PR by thinking it was a mistake).

If there's interest in it, we could add a "deterministic" flag similar to what we do now to show a contract is "Upgradeable."

# Changes

- Remove sentence about determinism
- Add links to the block explorer for each network.

How the docs look like after this PR

# How to test

Check that the links work as expected. No mention of any kind of determinism should be there.


anxolin (approved):
I'm good with it. It was misspell anyways 🙃

cowmarketing (approved):

PR #477, opened 2025-04-22
Bump contributor-assistant/github-action from 2.2.1 to 2.6.1  and remove org token

harisang:
This PR bumps contributor-assistant/github-action from 2.2.1 to 2.6.1 and removes the ORG_TOKEN, as i believe it is no longer needed


.github/workflows/cla.yml
```diff
@@ -15,7 +15,6 @@ jobs:
         uses: contributor-assistant/github-action@v2.2.1
```
fleupold: Let's bump to the latest version please.

fleupold (approved):

.github/workflows/cla.yml
harisang: Done

PR #498, opened 2025-05-26
Update competition_rules.md

acanidio-econ:
I removed the description of the batch auction and instead now discuss the fair combinatorial auction.

Note the following:
- we didn't have a specific page where we described the batch auction, and I kept this same structure. However, given that the fair combinatorial auction is more complex,  we may want to consider creating a dedicated page for it
- I don't explain the specific algorithm we use to select winners, and instead, I  provide a link to the CIP. In my opinion, we could discuss the winner selection algorithm if we introduce a specific page in the docs dedicated to the fair combinatorial auction but not in the current structure. 
- I kept the discussion about EBBO since, technically, it is still a rule

# Description

# Changes

- [ ] ...
- [ ] ...

## Summary by CodeRabbit

- **Documentation**
  - Updated the competition rules to introduce "Uniform Directional Clearing Prices (UDCP)" with new guidelines for order pricing and exceptions for certain orders.
  - Expanded the description of the auction process, including details on scoring, ranking, bid selection, and reward mechanisms.
  - Clarified rules on deadlines and penalties for winning solvers.


pretf00d (approved):
looks good

docs/cow-protocol/reference/core/auctions/competition_rules.md
```diff
@@ -18,11 +18,9 @@ All solvers participating in the solver competition must abide by certain rules.
...
...utions (and corresponding winning solvers) is chosen using a Fair Combinatorial Auction (see [CIP-67](https://snapshot.box/#/s:cow.eth/proposal/0xf9ecb08c4738f04c4525373d6b78085d16635f86adacd1b8ea77b2c176c99d32) At a high level, the protocol first considers bids containing only orders on the same token pair and in the same direction and computes the best bid on each directed token pair, where the best bid is the one generating the highest score. The collection of these best bids is the reference outcome for each directed token pair, representing the best possible execution against outside liquidity. The protocol then uses the reference outcome to filter out the remaining batched bids (batched because, by definition, they contain orders on multiple directed token pairs). A batched bid is filtered out whenever it generates lower score for a given directed token pair than the reference outcome. In the final step, the protocol considers all the batched bids that survived the filtering and the best bids on individual directed token pairs. It computes the collection of winning bids, under the constraint that all orders on the same directed token pair must be part of the same winning bid.
```
harisang:
```suggestion
- Every valid solution is associated with a score relative to the amount of surplus it generates for the users, as described in [CIP-38](https://snapshot.box/#/s:cow.eth/proposal/0xfb81daea9be89f4f1c251d53fd9d1481129b97c6f38caaddc42af7f3ce5a52ec) and [CIP-65](https://snapshot.box/#/s:cow.eth/proposal/0xd172281444c48254398881c57a57a2acbf0802a385e6c94384fd358b943aa4f4). The set of winning solutions (and corresponding winning solvers) is chosen using a Fair Combinatorial Auction (see [CIP-67](https://snapshot.box/#/s:cow.eth/proposal/0xf9ecb08c4738f04c4525373d6b78085d16635f86adacd1b8ea77b2c176c99d32). At a high level, the protocol first considers bids containing only orders on the same token pair and in the same direction and computes the best bid on each directed token pair, where the best bid is the one generating the highest score. The collection of these best bids is the reference outcome for each directed token pair, representing the best possible execution against outside liquidity. The protocol then uses the reference outcome to filter out the remaining batched bids (batched because, by definition, they contain orders on multiple directed token pairs). A batched bid is filtered out whenever it generates lower score for a given directed token pair than the reference outcome. In the final step, the protocol considers all the batched bids that survived the filtering and the best bids on individual directed token pairs. It computes the collection of winning bids, under the constraint that all orders on the same directed token pair must be part of the same winning bid. 
```

harisang (approved):

PR #559, opened 2025-11-21
Update the_problem.md

acanidio-econ:
Added an intro to explain what is the purpose of this section.

Added "cow amm orders" as order type

General rewrite

## Summary by CodeRabbit

* **Documentation**
  * Reframed the optimization problem as a two-step routing + bidding flow and expanded the solver description.
  * Replaced "batches" with "auctions" and clarified validity, execution cadence, and horizon language.
  * Renamed order taxonomy to Sell/Buy with explicit fill‑or‑kill and partial‑fill semantics.
  * Added CoW AMM order definitions, perpetual-auction behavior, and unified their scoring with sell orders.
  * Clarified solver vs. protocol fee wording and updated feasibility, scoring, and enforcement notes.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/docs/cow-protocol/reference/core/auctions/the_problem.md
+++ b/docs/cow-protocol/reference/core/auctions/the_problem.md
@@ -7,7 +7,7 @@ sidebar_position: 1
 
 At a high level, a solver is an algorithm that takes as inputs the orders that are valid for an auction, the state of the liquidity sources that the solver can access, the rules set by the protocol (including those specifying the protocol fees), and returns one or multiple _solutions_, where a solution selects a subset of the orders valid in an auction and specifies feasible an in and out amounts for each of those orders. The solutions proposed by each solver during a given auction are the outcome of a two-step optimization problem: optimal routing and then optimal bidding. 
 
-In this section, we describe mathematically the different components of the "solving" problem. We do not discuss optimal routing but we do discuss optimal bidding [here](rewards#solvers-strategy).
+In this section, we describe mathematically the different components of the "solving" problem. We  discuss optimal routing and optimal bidding [in the solvers' strategy section](rewards#solvers-strategy).
 
 ## User orders
 
--- a/docs/cow-protocol/reference/core/auctions/the_problem.md
+++ b/docs/cow-protocol/reference/core/auctions/the_problem.md
@@ -86,7 +86,7 @@ Solvers are also expected to charge a fee to cover the costs of executing an ord
 
 ## Solution
 
-Solvers propose solutions to the protocol, where a solution is a set of trades to execute. Formally, suppose there are $$I$$ users and CoW AMM orders and _J_ are external liquidity sources. A solution is a list of trades $$\{o_1, o_2, ...o_I, l_1, l_2, ..., l_J\}$$ for each user, CoW AMM pool and external liquidity source such that:
+Solvers propose solutions to the protocol, where a solution is a set of trades to execute. Formally, suppose there are $$I$$ users and CoW AMM orders and _J_ external liquidity sources. A solution is a list of trades $$\{o_1, o_2, ...o_I, l_1, l_2, ..., l_J\}$$ for each user, CoW AMM pool and external liquidity source such that:
 
 * **Incentive compatibility and feasibility**: the solution respects the orders' acceptance set.  
 * **Uniform directional clearing prices**: all users trading the same token pair in the same direction must face the same prices. Importantly, this constraint is defined at the moment when the swap occurs. So, for example, suppose user _i_ receives _x_ units of token 1 in exchange for _y_ units of token 2 and that the protocol takes a fee in the sell token $$f_2$$. Define $$p_{1,2}=\frac{y-f_2}{x}$$ as the price at which the swap occurs. Uniform directional clearing prices means that $$p_{1,2}$$ is the same for all users buying token 1 and selling token 2. Deviations from uniform directional prices are allowed to account for the extra gas cost of orders triggering hooks.
--- a/docs/cow-protocol/reference/core/auctions/the_problem.md
+++ b/docs/cow-protocol/reference/core/auctions/the_problem.md
@@ -80,7 +80,7 @@ Each user order may have an associated fee paid to the protocol. At a high level
 
 :::note
 
-Solvers are also expected to also charge a fee to cover the costs of executing an order (for example, gas). We discuss such fees later in the context of solvers' optimal bidding, but we do not account for them here as they are not part of the protocol.
+Solvers are also expected to charge a fee to cover the costs of executing an order (for example, gas). We discuss such fees later in the context of solvers' optimal bidding, but we do not account for them here as they are not part of the protocol.
 
 :::
 
--- a/docs/cow-protocol/reference/core/auctions/the_problem.md
+++ b/docs/cow-protocol/reference/core/auctions/the_problem.md
@@ -3,9 +3,11 @@ id: the-problem
 sidebar_position: 1
 ---
 
-# Optimization problem
+# What is solving? A matehmatical, high-level description
 
-In this section, we describe all the different components of the optimization problem that needs to be solved within each batch.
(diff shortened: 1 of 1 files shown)


docs/cow-protocol/reference/core/auctions/the_problem.md
fhenneke, line 22: This renders a bit strange as _surplus_function_. Maybe "... we define the surplus function ..."?

docs/cow-protocol/reference/core/auctions/the_problem.md
```diff
@@ -53,26 +57,40 @@ Again, the surplus function is defined as
...
 
-Solvers are also expected to charge a fee to cover the cost of executing an order. We discuss such fees later in the context of solvers' optimal bidding, but we do not account for them here as they are not part of the protocol.
+Solvers are also expected to charge a fee to cover the costs of executing an order (for example, gas). We discuss such fees later in the context of solvers' optimal bidding, but we do not account for them here as they are not part of the protocol.
 
 :::
 
 ## Solution
 
-Solvers propose solutions to the protocol, where a solution is a set of trades to execute. Formally, suppose there are $$I$$ users and _J_ liquidity sources. A solution is a list of trades $$\{o_1, o_2, ...o_I, l_1, l_2, ..., l_J\}$$ one per user and one per liquidity source such that:
+Solvers propose solutions to the protocol, where a solution is a set of trades to execute. Formally, suppose there are $$I$$ users and CoW AMM orders and _J_ are external liquidity sources. A solution is a list of trades $$\{o_1, o_2, ...o_I, l_1, l_2, ..., l_J\}$$ for each user, CoW AMM pool and external liquidity source such that:
```
fhenneke:
```suggestion
Solvers propose solutions to the protocol, where a solution is a set of trades to execute. Formally, suppose there are $$I$$ users and CoW AMM orders and _J_ external liquidity sources. A solution is a list of trades $$\{o_1, o_2, ...o_I, l_1, l_2, ..., l_J\}$$ for each user, CoW AMM pool and external liquidity source such that:
```

docs/cow-protocol/reference/core/auctions/the_problem.md
fhenneke, line 14: I still do not see the advantage of identifying orders with acceptance sets. For example there are orders with the same acceptance set but different protocol fees.

docs/cow-protocol/reference/core/auctions/the_problem.md
```diff
@@ -21,9 +23,11 @@ We also assume that $$0 \in S$$ that is, when submitting an order a user accepts
...
 
-A _limit sell order_ specifies a maximum sell amount of a given token _Y_ > 0, a buy token _b_, and a limit price $$\pi$$, that corresponds to the worst-case exchange rate that the user is willing to settle for. They can be fill-or-kill whenever the executed sell amount must be Y (or nothing). They can be partially fillable if the executed sell amount can be smaller or equal to Y.  Formally, if _x_ denotes the (proposed) buy amount and _y_ denotes the (proposed) sell amount of the order, a fill-or-kill limit sell order has the form
+Sell orders can be fill-or-kill whenever the executed sell amount must be Y (or nothing). They can be partially fillable if the executed sell amount can be smaller or equal to Y.  Formally, if _x_ denotes the (proposed) buy amount and _y_ denotes the (proposed) sell amount of the order, a fill-or-kill limit sell order has the form
```
fhenneke: This reads a bit confusing to me. Maybe something like
```suggestion
Sell orders can be fill-or-kill which means that the executed sell amount must be Y (or nothing). They can be partially fillable which means that the executed sell amount can be smaller or equal to Y.  Formally, if _x_ denotes the (proposed) buy amount and _y_ denotes the (proposed) sell amount of the order, a fill-or-kill limit sell order has the form
```
Or maybe "Sell orders are called fill-or-kill whenever the executed sell amount must be Y (or nothing). ..."

docs/cow-protocol/reference/core/auctions/the_problem.md
```diff
@@ -83,7 +101,7 @@ At CoW DAO's discretion, systematic violation of these rules may lead to penaliz
 
 From the protocol viewpoint, each solution that satisfies the above constraints has a _score_ that is given by the total surplus generated and the fees paid to the protocol, all aggregated and denominated in some numéraire. More specifically, the score of a solution is equal to the sum of scores of the orders the solution proposes to execute, where the score of an order $$o$$ is defined as:
 
-* $$o$$ is a sell order: $$\mathrm{score}(o) = (U(o)+ f(o)) \cdot p(b)$$, where $$p(b)$$ is an externally provided price of the buy token relative to a numéraire.
+* $$o$$ is a sell order or a CoW AMM order: $$\mathrm{score}(o) = (U(o)+ f(o)) \cdot p(b)$$, where $$p(b)$$ is an externally provided price of the buy token relative to a numéraire.
```
fhenneke, line 104: This is the intention, I guess, so we should keep this change.

But we should probably check internally if we want to enforce it. At the moment, CoW AMM orders can also be buy orders (and solvers can choose buy or sell orders based on whichever is best for them.

fhenneke (approved):
Just left a few nitpicks.

Generally reads a lot better than before. Adding CoW AMMs explicitly is also a good idea.

docs/cow-protocol/reference/core/auctions/the_problem.md
acanidio-econ: Acceptance sets are defined from the users' perspective. It is fine if the orders have the same acceptance set but different fees: it just means that the "in/out net of fee" that a user is willing to accept is the same as that of another user

docs/cow-protocol/reference/core/auctions/the_problem.md
acanidio-econ: right, I didn't know that. Then there are two types of CoW AMM orders: CoW AMM buy and CoW AMM sell orders

docs/cow-protocol/reference/core/auctions/the_problem.md
acanidio-econ: I don't know if you can have a general definition of surplus function that could apply to any order (basket order or whatever). I guess beside saying that it should be increasing in the buy amounts and decreasing in the sell amounts, I don't know if you can say more *in general*. Of course, once you have a specific order type, you can specify the surplus function for that order type

fhenneke: I see how the acceptance set is net of fees and a valid execution only depends on that set. But that set is not enough to encode all information of the order required for the mechanism.
So my gripe is more with concepts below which are defined in terms of the set but should be defined in terms of the order.

For example, two different orders can have the same acceptance set but different protocol fees. The protocol fee is not really an object `f_S` with an order `S`. Instead, there is an order `o` with an acceptance set `S_o` and a protocol fee function `f_o`.

None of that needs to be addressed in this PR though.

docs/cow-protocol/reference/core/auctions/the_problem.md
```diff
@@ -3,9 +3,11 @@ id: the-problem
 sidebar_position: 1
 ---
 
-# Optimization problem
+# What is solving? A matehmatical, high-level description
```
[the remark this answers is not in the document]
fhenneke: The typo seems not to be fixed yet.

docs/cow-protocol/reference/core/auctions/the_problem.md
fhenneke: I was trying to refer to the formatting of `... we define _surplus_function_ $$U_S:S\rightarrow \mathbb R$$, ...`. I see how in general surplus is not easy to define. One could leave out `_surplus_function_` or change it to `... the utility function ...`.

docs/cow-protocol/reference/core/auctions/the_problem.md
fhenneke: I think it is fine to only explain the sell order case for now, and we should encourage solvers to only use sell orders for consistency.

docs/cow-protocol/reference/core/auctions/the_problem.md
fhenneke:
```suggestion
# What is solving? A mathematical, high-level description
```

pretf00d (approved):

PR #595, opened 2026-02-12
docs: add comprehensive generalized wrappers documentation

kaze-cow:
# Description

Add complete documentation for CoW Protocol's generalized wrappers feature across three audience-specific files:

# Changes

- [x] `concepts/order-types/wrappers.md`: High-level overview explaining what wrappers are, use cases (Euler leverage, flash loans, TWAP, protocol hooks), benefits, and considerations
- [x] `integrate/wrappers.mdx`: Practical integration guide with three sections:
  * For Order Creators: How to add wrappers to orders via appData
  * For Wrapper Developers: Building custom wrapper contracts with security requirements and implementation examples
  * For Solvers: Encoding wrapper settlements and handling wrapper execution
- [x] `reference/contracts/periphery/wrapper.mdx`: Smart contract reference focused on ICowWrapper interface, CowWrapper abstract contract, implementation patterns, and on-chain behavior

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive guides for Generalized Wrappers: concepts, architecture, lifecycle, and quick-start examples.
  * New integration guide for building, encoding, chaining, validation, solver behavior, execution/authentication, and troubleshooting.
  * Reference material on interfaces and flow, data-format specifications, on-chain utilities, testing guidance, security considerations, gas trade-offs, and example implementations.


docs/cow-protocol/concepts/order-types/wrappers.md
```diff
@@ -0,0 +1,99 @@
...
+
+## What are Wrappers?
+
+Wrappers are smart contracts that "wrap" the settlement process, executing custom logic surrounding settlement contract. When a solver executes a settlement that includes a wrapper, they call the wrapper contract instead of the settlement contract directly. The wrapper calls the settlement contract on behalf of the solver.
+
+This mechanism extends CoW Protocol's functionality in a modular way, allowing new features and integrations to be added without modifying the core settlement contract or requiring any changes to solver implementations.
+
+### Wrapper Authentication
+
+For security, all wrappers must be approved through CoW Protocol's `GPv2AllowlistAuthenticator` before they can be used. This ensures that only audited and  wrappers can interact with the settlement contract, protecting users, solver, and the protocol from malicious contracts. Unlike [hooks](./cow-hooks.mdx), Wrapper functionality can revert the execution of a transaction, ensuring sensitive user operations such as cross-chain operations can be guarenteed to be completed.
```
pretf00d: This line doesn't make sense; "This ensures that only audited and  wrappers" also double space after "and"

pretf00d: solver ---> solvers

pretf00d: title for Wrapper or lower case wrapper?

pretf00d: cross-chain operations can be guarenteed to be completed. --> can be completed safely.

docs/cow-protocol/concepts/order-types/wrappers.md
```diff
@@ -0,0 +1,99 @@
...
+
+### Protocol-Approved Hooks
+
+Unlike [CoW Hooks](./cow-hooks.mdx), which can revert even if the order is executed successfully, wrappers provide a way to enforce required pre- and post-settlement operations. Since wrappers are protocol-approved through the allowlist authenticator, they can implement critical functionality that must execute:
+
+- Compliance checks (e.g., OFAC screening)
+- Cross chain transfers (pre- or post- transfer)
+- Deposit in a vault or other wrapper contract (swap and stake)
+
+The key difference from hooks is that if a wrapper is required for an order, the settlement cannot proceed without it—making wrappers ideal for functionality that must not be skipped that is also approved.
```
pretf00d: making wrappers ideal for functionality that must not be skipped that is also approved. ---> making wrappers ideal for transactions requiring more robust checks.

docs/cow-protocol/concepts/order-types/wrappers.md
```diff
@@ -0,0 +1,99 @@
...
+
+- The complexity of the wrapper's logic
+- The size of the settlement data being passed through
+- The number of nested wrappers in the chain
+
+For many use cases, this overhead is acceptable given the functionality unlocked, but it's an important factor in deciding whether to use a wrapper.
+
+### Requires Protocol Approval
+
+Wrappers cannot be deployed and used immediately—they must be approved by the CoW DAO through the allowlist authenticator. This approval process ensures high quality wrapper implementations and safety for solvers, but means there's a roadblock for developers looking to extend CoW Protocol. Developers should plan for this approval process when building wrapper-based integrations.
```
pretf00d: they must be approved by the CoW DAO through the allowlist authenticator --> they must be approved by CoW DAO through the allowlist authenticator.

docs/cow-protocol/integrate/wrappers.mdx
```diff
@@ -0,0 +1,346 @@
...
+
+To build a wrapper, you will:
+
+1. Inherit from the `CowWrapper` abstract contract
+2. Implement `_wrap()` with your pre/post-settlement logic
+3. Implement `parseWrapperData()` for input validation
+4. Test thoroughly on testnets
+5. Submit for allowlist approval by the CoW Protocol team
+
+Read up on the details of creating a wrapper from [the wrapper contract information page](../reference/contracts/periphery/wrapper.mdx).
```
pretf00d: Read up on the details of creating a wrapper from [the wrapper --> Read more about the details of creating a wrapper from [the wrapper

docs/cow-protocol/integrate/wrappers.mdx
```diff
@@ -0,0 +1,346 @@
...
+
+---
+
+## For Solvers
+
+This section explains how to execute settlements that include wrapper contracts as part of your solver implementation.
+
+## Detecting Wrapper Orders
+
+Wrappers are specified in the order's [`appData`](../reference/core/auctions/schema.md) under the `wrappers` field:
```
pretf00d, line 143: order's ---> orders

docs/cow-protocol/integrate/wrappers.mdx
```diff
@@ -0,0 +1,346 @@
...
+
+If `"is_omittable": true`, you MAY skip the wrapper if you find a better solution without it.
+
+### 2. Verify Wrapper Authentication
+
+All approved wrappers will be approved by the DAO and registered in [`GPv2AllowlistAuthenticator`](../reference/contracts/core/allowlist.md). It is recommended to verify wrappers are allowlisted before including them in settlements.
+
+### 3. Include Wrappers in Solution
+
+Whether using the CoW-provided driver or your own, ensure wrapper information flows from the auction to your final settlement transaction.
```
pretf00d: CoW-provided driver or your own --> CoW can't be on its own, needs to be CoW DAO or CoW Protocol

docs/cow-protocol/integrate/wrappers.mdx
```diff
@@ -0,0 +1,346 @@
...
+
+## Encoding Wrapper Settlements
+
+:::info
+If you're using the driver provided by the CoW team, you can skip this section - the driver handles encoding automatically.
```
pretf00d: As above no single use of CoW

docs/cow-protocol/integrate/wrappers.mdx
```diff
@@ -0,0 +1,346 @@
...
+In addition to producing the correct encoded data, the helper also validates:
+- All wrappers are allowlisted
+- All wrappers use the same settlement contract
+- Each wrapper can parse its data
+
+## Accumulating Wrappers in Solutions
+
+### Using CoW-Provided Driver
+
+If you're using the CoW team's driver, wrappers are handled automatically. Just ensure your solver process includes wrapper information in the solution output.
```
pretf00d: as above, no single 'CoW'

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
```diff
@@ -0,0 +1,497 @@
...
+### Key Design Principles
+
+1. **Abstract**: There are very few limitations on what a wrapper can/can't do around a settlement transaction
+2. **Efficient Encoding**: Wrapper-specific data appended to minimize gas overhead
+3. **Nested Support**: Multiple wrappers chain by encoding addresses sequentially, allowing CoW orders
+4. **Authentication**: Only allowlisted wrappers can call settlement contract
+
+## Implementation
+
+Developers looking to integrate using wrappers should copy this all-in-one solidity file in your project.
```
pretf00d: in your project --> into their project folders.

pretf00d (changes requested):
Left comments

docs/cow-protocol/integrate/wrappers.mdx
kaze-cow: I believe the grammar is it is now is correct here. The appData is "possessed" by the order.

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
kaze-cow: aba6279

docs/cow-protocol/integrate/wrappers.mdx
kaze-cow: 3e58601

docs/cow-protocol/integrate/wrappers.mdx
kaze-cow: 3e58601

docs/cow-protocol/integrate/wrappers.mdx
kaze-cow: ad8f9e6

docs/cow-protocol/integrate/wrappers.mdx
kaze-cow: 5563ab8

docs/cow-protocol/concepts/order-types/wrappers.md
kaze-cow: 5563ab8

docs/cow-protocol/concepts/order-types/wrappers.md
kaze-cow: good wording, made small adjustments myself 44a2d3b

docs/cow-protocol/concepts/order-types/wrappers.md
kaze-cow: also included in rewording 8e0fcdd

kaze-cow: should have been lower case. also fixed in 8e0fcdd

kaze-cow: included in 8e0fcdd

kaze-cow: I agree it was a bit confusing. Reworded and reordered. 8e0fcdd

docs/cow-protocol/concepts/order-types/wrappers.md
```diff
@@ -0,0 +1,91 @@
...
+
+## What are Wrappers?
+
+Wrappers are smart contracts that "wrap" the settlement process, executing custom logic surrounding settlement contract. When a solver executes a settlement that includes a wrapper, they call the wrapper contract instead of the settlement contract directly. The wrapper calls the settlement contract on behalf of the solver.
+
+This mechanism extends CoW Protocol's functionality in a modular way, allowing new features and integrations to be added without modifying the core settlement contract or requiring any changes to solver implementations.
+
+## Use Cases
+
+Wrappers enable a wide variety of advanced trading and DeFi operations:
```
fedgiac: We should make it clear that these are hypothetical and there isn't a corresponding feature out. Overall, I get the feeling this section should be much shorter since most of the people don't need a detailed description of hypothetical use cases, a bullet list would be enough. Later, we can link here to the concrete Euler implementation.

docs/cow-protocol/concepts/order-types/wrappers.md
```diff
@@ -0,0 +1,91 @@
...
+
+The key difference from hooks is that if a wrapper is required for an order, the settlement cannot proceed without it—making wrappers ideal for transactions requiring robust execution.
+
+## Considerations
+
+While wrappers are powerful, there are important considerations to keep in mind:
+
+### Gas Overhead
+
+Wrappers add gas overhead to settlement transactions. Benchmarks show that even an empty wrapper (one that does nothing but pass through to settlement) adds approximately 22,272 gas, or about 11.4% overhead. The actual overhead depends on:
```
fedgiac: Huh, that's a lot, I expect much less for an empty wrapper. Sure, we can get any number here if the calldata is large enough. How was the number computed?
Anyway, I think we should try to be more specific here. It would be nice to try a few different settlement data size, which should be the only real parameter in the case of an empty wrapper. (Something like [this](https://nomevlabs.slack.com/archives/C043PJXF657/p1771960023624619?thread_ts=1771585692.165389&cid=C043PJXF657).)

docs/cow-protocol/concepts/order-types/wrappers.md
```diff
@@ -0,0 +1,91 @@
...
+
+- The complexity of the wrapper's logic
+- The size of the settlement data being passed through
+- The number of nested wrappers in the chain
+
+For many use cases, this overhead is acceptable given the functionality unlocked, but it's an important factor in deciding whether to use a wrapper.
+
+### Requires Protocol Approval
+
+Wrappers cannot be deployed and used immediately—they must be approved by CoW DAO through the allowlist authenticator. This approval process ensures high-quality wrapper implementations and safety for solvers, but means there's a roadblock for developers looking to extend CoW Protocol. Developers should plan for this approval process when building wrapper-based integrations.
```
fedgiac, line 38: Not for this PR, but eventually we should add the minimal things to check to be considered for approval.

docs/cow-protocol/concepts/order-types/wrappers.md
fedgiac: Ok, maybe this is too specific for this very general page, but then I'd remove the concrete gas amount and do a more detailed analysis in another page.

docs/cow-protocol/concepts/order-types/wrappers.md
```diff
@@ -0,0 +1,91 @@
...
+
+Wrappers are a powerful tool for advanced integrations on CoW Protocol. To start building with wrappers:
+
+- **For developers**: See the [Integration Guide](../../integrate/wrappers.mdx) for implementation details, code examples, and security guidelines
+- **For technical specs**: Consult the [Technical Reference](../../reference/contracts/periphery/wrapper.mdx) for detailed contract documentation and API specifications
+
+To learn more about wrappers and see example implementations:
+
+- [Euler Integration Contracts Repository](https://github.com/cowprotocol/euler-integration-contracts) - Contains the `CowWrapper` abstract contract and example implementations
+- [Services Repository PR #3700](https://github.com/cowprotocol/services/pull/3700) - Backend integration implementation. Good reference for solvers looking to support wrappers.
```
fedgiac: To be extra clear, I'd mention that this isn't needed for people looking to implement wrappers.

docs/cow-protocol/integrate/wrappers.mdx
```diff
@@ -0,0 +1,346 @@
...
+
+### Overview
+
+To build a wrapper, you will:
+
+1. Inherit from the `CowWrapper` abstract contract
+2. Implement `_wrap()` with your pre/post-settlement logic
+3. Implement `parseWrapperData()` for input validation
+4. Test thoroughly on testnets
+5. Submit for allowlist approval by the CoW Protocol team
```
fedgiac: I'm worried about this, the last thing we want to do is wading through a bunch of half-assed wrapper implementations. We should really add some details on this, the bare minimum is a contract audit from a well-known provider.

docs/cow-protocol/integrate/wrappers.mdx
```diff
@@ -0,0 +1,346 @@
...
+
+To use a wrapper in your order, add it to the `appData` when creating the order:
+
+### AppData Structure
+
+```typescript
+interface WrapperCall {
+    target: string;        // Wrapper contract address (must be allowlisted)
+    data?: string;         // Optional hex-encoded wrapper-specific data
+    is_omittable?: boolean; // Whether solver can skip this wrapper (default: false)
```
fedgiac: I'm not up to speed with this. Is this how it works now? I thought we decided to not use `is_omittable` in the end.
Also, is it right that it's snake_case and not camelCase?

docs/cow-protocol/integrate/wrappers.mdx
```diff
@@ -0,0 +1,346 @@
...
+        }
+    ]
+}
+```
+
+The wrappers execute in sequence: Wrapper1 → Wrapper2 → Settlement → Wrapper2 (post) → Wrapper1 (post). Note that wrappers from other users' orders may be interspersed, though this should generally not affect the execution of the order.
+
+### Validating Wrapper Configuration
+
+Before submitting an order, you can confirm the wrapper encoding is valid using `CowWrapperHelper`:
```
fedgiac: Here it needs a link to the helper address, or some way to recover it.

docs/cow-protocol/integrate/wrappers.mdx
```diff
@@ -0,0 +1,346 @@
...
+for (const order of auction.orders) {
+    if (order.appData.wrappers) {
+        // Add wrappers needed for this order
+        solution.wrappers.push(...order.appData.wrappers);
+    }
+}
+
+// Encode final transaction
+const tx = encodeWrapperSettlement(solution);
+```
```
fedgiac, line 294: All the magic is hidden in `encodeWrapperSettlement`. I think it's fair to assume that the solvers know how to create the settle calldata and use that together with `verifyAndBuildWrapperData`.

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
```diff
@@ -0,0 +1,705 @@
...
+sequenceDiagram
+    participant Solver
+    participant Wrapper1
+    participant Wrapper2
+    participant Settlement
+
+    Solver->>Wrapper1: wrappedSettle(settleData, wrapperData)
+    activate Wrapper1
+    Note over Wrapper1: Pre-settlement logic
+    Wrapper1->>Wrapper2: _next(remainingData)
```
fedgiac: Let's show here that `settleData` should be passed along.

docs/cow-protocol/concepts/order-types/wrappers.md
```diff
@@ -0,0 +1,91 @@
...
+
+Wrappers cannot be deployed and used immediately—they must be approved by CoW DAO through the allowlist authenticator. This approval process ensures high-quality wrapper implementations and safety for solvers, but means there's a roadblock for developers looking to extend CoW Protocol. Developers should plan for this approval process when building wrapper-based integrations.
+
+### On-Chain Protocol Does Not Enforce Execution
+
+Despite official rules enforcing the execution of wrappers when required by the user, a solver may choose not to execute a wrapper with an order. This means wrappers must be designed defensively:
+
+- If a wrapper is strictly required, the order should fail to settle without it
+- Wrappers should validate all input data and fail in cases where a user's funds could be at risk
+
```
fedgiac, line 47: Needed section: something like

```
### On-Chain Protocol Does Not Validate Settlement
```

where we say that a malicious solver can change the settlement data and that a wrapper should not trust that to be executed.

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
```diff
@@ -0,0 +1,705 @@
...
+## Implementation
+
+Developers looking to integrate using wrappers should copy this all-in-one solidity file into their project as vendored dependency.
+
+It provides a few base contracts which can serve as the foundation for integration:
+
+- **`ICowWrapper`**: Core interface all wrappers must implement
+- **`CowWrapper`**: Abstract base contract providing security and utilities that all wrappers should use
+
+```solidity
```
fedgiac: I'd remove this and add a link to the Euler project, so we're sure it's kept up-to-date.

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
```diff
@@ -0,0 +1,705 @@
...
+            // More wrapper data remains - call the next wrapper in the chain
+            CowWrapper(nextWrapper).wrappedSettle(settleData, remainingWrapperData);
+        }
+    }
+}
+```
+
+### Quick Start Example
+
+Here's a minimal wrapper implementation to use as a starting point:
```
fedgiac: Not for this PR, but: it would be extra cool if we were to provide a template repo with just this.

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
```diff
@@ -0,0 +1,705 @@
...
+**Parameters:**
+- `settleData`: Original `GPv2Settlement.settle(...)` calldata
+- `chainedWrapperData`: 
+
+#### Chained Wrapper Data Encoding
+
+Each wrapper in the chain consists of:
+- 2 bytes: uint16 length of wrapper-specific data
+- `length` bytes: wrapper-specific data for this wrapper
+- 20 bytes: address of next wrapper (omitted for the final wrapper)
```
fedgiac: It's confusing because the actual final wrapper should be included).

```suggestion
- 20 bytes: address of next wrapper (or stop here if no wrappers are left)
```

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
```diff
@@ -0,0 +1,705 @@
...
+The final wrapper in the chain omits the next wrapper address and calls SETTLEMENT directly.
+
```
fedgiac: It's confusing because the actual final wrapper should be included).

```suggestion
- 20 bytes: address of next wrapper (or stop without an address if no wrappers are left)

When no wrapper is left, the final wrapper in the chain calls SETTLEMENT directly.
```

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
```diff
@@ -0,0 +1,705 @@
...
+Example:
+```text
+[0x0005][0xAABBCCDDEE][0x1234...ABCD][0x0003][0x112233]
+↑len    ↑data         ↑next wrapper  ↑len   ↑data
```
fedgiac: For extra clarity, let's spell out the address. Also, spacing typo.
```suggestion
[0x0005][0xAABBCCDDEE][0x1234567890123456789012345678901234567890][0x0003][0x112233]
↑len    ↑data         ↑next wrapper                               ↑len    ↑data
```

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
```diff
@@ -0,0 +1,705 @@
...
+Wrapper data uses an optimized encoding to minimize gas overhead:
+
+### Structure
+
+```
+┌─────────────────────┬──────┬──────────┬──────────┬──────┬──────────┐
+│   Settle Calldata   │ Len₁ │  Data₁   │  Addr₂   │ Len₂ │  Data₂   │
+│   (to settlement)   │(2 B) │ (wrap1)  │ (20 B)   │(2 B) │ (wrap2)  │
+└─────────────────────┴──────┴──────────┴──────────┴──────┴──────────┘
+│<─── settleData ───>│<──────────── wrapperData ─────────────────────>│
```
fedgiac: Typo in spacing.

```suggestion
└─────────────────────┴──────┴──────────┴──────────┴──────┴──────────┘
│<─── settleData ────>│<─────────── wrapperData ────────────────────>│
```

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
```diff
@@ -0,0 +1,705 @@
...
+
+**Components:**
+- **Settle Calldata**: `GPv2Settlement.settle(...)` calldata (function selector + args)
+- **Len₁**: 2-byte (uint16) length of wrapper 1's data
+- **Data₁**: Wrapper 1's custom data
+- **Addr₂**: 20-byte address of wrapper 2
+- **Len₂**: 2-byte (uint16) length of wrapper 2's data
+- **Data₂**: Wrapper 2's custom data
+
+Note: The settlement contract address is derived from the `SETTLEMENT` immutable and is not included in `wrapperData`.
```
fedgiac: Wait what are we encoding here? We discussed the wrapper encoding, that's it, `settle` doesn't need to be injected. All of this shouldn't be here right?

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
```diff
@@ -0,0 +1,705 @@
...
+
+### Data Consumption
```
fedgiac: Repeating information?

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
```diff
@@ -0,0 +1,705 @@
...
+    ) external view returns (bytes memory wrapperData);
+}
+```
+
+
+### `CowWrapperHelpers`
+
+To aid external integrators in encoding and verifying wrapper data, an additional view-only contract is provided.
+
+See the [CowWrapperHelper contract source code](https://github.com/cowprotocol/euler-integration-contracts/blob/main/src/CowWrapperHelper.sol) for the complete implementation.
```
fedgiac: Broken link.

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
```diff
@@ -0,0 +1,705 @@
...
+contract CowWrapperHelper {
+    /// @notice Validate and build wrapper data for multiple wrappers
+    /// @param wrapperAddresses Array of wrapper contract addresses
+    /// @param wrapperDatas Array of wrapper-specific data (parallel to addresses)
+    /// @return wrapperData Encoded wrapper data ready for wrappedSettle call
+    function verifyAndBuildWrapperData(
+        address[] calldata wrapperAddresses,
+        bytes[] calldata wrapperDatas
+    ) external view returns (bytes memory wrapperData);
+}
```
fedgiac: Incorrect function signature.

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
```diff
@@ -0,0 +1,705 @@
...
+
+
+#### `validateWrapperData`
+
+Validates wrapper-specific data. Must be deterministic and revert on invalid input.
+
+```solidity
+function validateWrapperData(
+    bytes calldata wrapperData
+) external view override returns (bytes memory);
```
fedgiac: Incorrect function signature again. Once is ok but something went wrong with writing this if this is a repeating issue. I'm guessing it's LLM generated, but in case it's still on you to make sure the docs match the actual code. I won't check other function signatures until you confirm you checked them yourself.

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
```diff
@@ -0,0 +1,705 @@
...
+Though a solver would be slashed for doing so, there is no hard guarantee wrapper executes even if user specifies it. If wrapper is critical:
+
+**Option A: Make Order Fail Without Wrapper**
+```solidity
+// Pull required funds in wrapper
+function _wrap(bytes calldata wrapperData) internal override {
+    vault.withdraw(user, token, amount);  // Funds now in wrapper
+    _next(wrapperData);
+    // Settlement uses these funds - fails if wrapper didn't execute
+}
```
fedgiac: This is terrible, the solver can just take the money from the user and not execute the order, leading to loss of funds. This should totally not be here.

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
```diff
@@ -0,0 +1,705 @@
...
+**Option B: EIP-1271 Authentication**
+```solidity
+// Order signature only valid if wrapper executed
+function isValidSignature(bytes32 hash, bytes memory signature)
+    external view returns (bytes4)
+{
+    require(wrapperExecuted[hash], "Wrapper required");
+    return EIP1271_MAGIC;
+}
+```
```
fedgiac: This is actually extremely annoying and complex to do because of the cyclical dependency on the app data. The wrapper can't generate `hash` without knowing the app data unless doing introspection on the settlement.

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
```diff
@@ -0,0 +1,705 @@
...
+    {
+        lendingPool = ILendingPool(_lendingPool);
+    }
+
+    function _wrap(bytes calldata settleData, bytes calldata wrapperData, bytes calldata remainingWrapperData) internal override {
+        (address asset, uint256 amount) =
+            abi.decode(wrapperData, (address, uint256));
+
+        // Store data for use in callback
+        // (In a real implementation, you would store settleData and remainingWrapperData)
```
fedgiac: Let's drop this section. It feels like this is the most important thing to show and it's omitted. Also, it probably requires some kind of authentication and is complicated to do, and I wouldn't want that we provide a known-broken implementation of a wrapper.

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
```diff
@@ -0,0 +1,705 @@
...
+## Known Limitations
+
+### Authentication Required
+
+Wrappers need protocol approval, and are not fully permissionless.
+
+### Settlement Upgrades
+
+When settlement contract upgrades, all wrappers must be redeployed. It is recommended to use [Cannon](https://usecannon.com/learn) to make it easy to redeploy your wrappers.
+
```
fedgiac: We already discussed limitations somewhere else, no need to do a much shorter versions of it. I also wouldn't mention upgrades like it was something we're happy to do willy-nilly.

```suggestion
```

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
```diff
@@ -0,0 +1,705 @@
...
+## Resources
+
+### Documentation
+- **[Wrapper Concepts](../../../concepts/order-types/wrappers.md)** - High-level overview and use cases
+- **[Integration Guide](../../../integrate/wrappers.mdx)** - Implementation guide for order creators and developers
+- **[GPv2Settlement](../core/settlement.md)** - Core settlement contract documentation
+
+### Code
+- **[CowWrapper.sol](https://github.com/cowprotocol/euler-integration-contracts/blob/main/src/CowWrapper.sol)** - Abstract contract and interfaces
```
fedgiac: Broken link.

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
```diff
@@ -0,0 +1,705 @@
...
+- **[EmptyWrapper](https://github.com/cowprotocol/euler-integration-contracts)** - Minimal reference implementation
+- **[Euler Integration](https://github.com/cowprotocol/euler-integration-contracts/pull/6)** - Complete production wrapper example
```
fedgiac: Incorrect links.

pretf00d (changes requested):
Waiting for what looks like it will be shortened before reviewing.

docs/cow-protocol/concepts/order-types/wrappers.md
kaze-cow: Makes sense. condensed to a short bullet point section, and worded the header so it alludes to the future expectations and not current reality 970be36

docs/cow-protocol/concepts/order-types/wrappers.md
kaze-cow: > Huh, that's a lot, I expect much less for an empty wrapper. Sure, we can get any number here if the calldata is large enough. How was the number computed?

I computed it by running a test settlement including a call through the `EmptyWrapper` using the playground in the services repo. [I first reported the wrapper usage here in the internal docs](https://www.notion.so/cownation/Generalized-Wrapper-2798da5f04ca8095a2d4c56b9d17134e?source=copy_link#27a8da5f04ca809ea96ac3eeb5b9e06d), and I have been updating it every time we make a significant change over time. I agree, its a bit higher than I expected. We can talk about the reasons why later.

I moved the discussion about gas overhead mostly into the integration guide and only touch on a high level here now. a57a461

docs/cow-protocol/concepts/order-types/wrappers.md
kaze-cow: yep, we [have this list internally](https://www.notion.so/cownation/Generalized-Wrapper-2798da5f04ca8095a2d4c56b9d17134e?source=copy_link#2858da5f04ca80afa7a8e2ee0595f6c4). Will expand upon this in future PR as you suggest.

docs/cow-protocol/concepts/order-types/wrappers.md
kaze-cow: added segments to the front of the sentance to make it clear who each resource is intended for 026faa2

docs/cow-protocol/integrate/wrappers.mdx
kaze-cow: I went ahead and put in our full list of implementation requirements and a link directly to it here 8efd607

docs/cow-protocol/integrate/wrappers.mdx
kaze-cow: >I'm not up to speed with this. Is this how it works now? I thought we decided to not use is_omittable in the end.

It is how it works, yes, and it is upstreamed in both the [`services`](https://github.com/cowprotocol/services/blob/82d3e7c5d72410f3438009a7a595c21f25382395/crates/app-data/src/app_data.rs#L78) and [`cow-sdk`](https://github.com/cowprotocol/cow-sdk/blob/main/packages/app-data/src/schemas/wrappers/v0.2.0.json#L25) repos.

Its snake case in rust due to rust code style, and camel case everywhere else. The integration guide should definitely be displaying the camelCase 99f2fc9

docs/cow-protocol/integrate/wrappers.mdx
kaze-cow: yea true. We haven't officially deployed this contract yet (my fault, sorry), so for right now [I put it in a gist](https://gist.github.com/kaze-cow/8c8a49f3305519948b1ec868c7efa5b9) and linked it here. We will need to come back here and update it when it becomes an official cow protocol deployment (probably in the next few weeks). e6d4223

docs/cow-protocol/integrate/wrappers.mdx
kaze-cow: definitely some clariifcation was needed here. Referred explicitly to the encoding section above and also showed the actual construction of the txn object dc0bf1d

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
kaze-cow: 0258e4c

docs/cow-protocol/concepts/order-types/wrappers.md
kaze-cow: Good idea. I tried to merge these two ideas together (since they are extremeley similar) as one section:

```
Wrapper and Settlement Call Pairing not Guarenteed On-chain
```

ab4864d

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
kaze-cow: I actually wanted to have an effective embed of the code contents here because its useful to see on the page, but unfortunately this doesn't seem to be easily supported by the current docs website markdown setup. I think it may be best to just manually update it. thoughts?

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
kaze-cow: good idea! Lets see how the 2nd usage of wrappers goes and we can see what the patterns are.

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
kaze-cow: Yea I think you are right. I was sort of thinking about it in terms of the call structure as a whole, but as you mention this isnt helpful for understanding the actual encoding pattern, so removing and clarifying what is displayed is for `chainedWrapperData`. 74bbdfa

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
kaze-cow: yea I guess its overkill with the existing docs about the behavior of the internal functions

kaze-cow: e163784

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
kaze-cow: a5fa98c

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
kaze-cow: updated call signatures 8726c82

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
kaze-cow: yea looking through all this it seems like it would be better to save these types of examples for later as a deliberate separate PR add

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
kaze-cow: sorry, this example should basically be combined with EIP-1271 (which is specified below). Overall I kind of feel like the examples should be removed entirely because its highly case-by-case basis issue, so I went ahead and removed them. 90e0cd6

(btw the cyclical issue really is annoying. I have a couple ways to deal with it, but so far they always involve effectively passing the entire order data to hash into the function. Not efficient, but its doable and we already do it for euler)

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
kaze-cow: 90e0cd6

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
kaze-cow: agreed, I have a proper solution I think for this, but lets save for another PR. For now removed this example entirely because its highly implementation/need dependent 90e0cd6

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
kaze-cow: fixed at same time as other previous link

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
kaze-cow: fixed 78b02f2

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
fedgiac: Maybe @anxolin can help here. I think at some point you were able to use code snippets in the docs for a code tutorial, how did you do it?

kaze-cow:
requesting re-review since there is only one outstanding comment that is pending non-blocking query

docs/cow-protocol/concepts/order-types/wrappers.md
```diff
@@ -0,0 +1,60 @@
...
+## What are Wrappers?
+
+Wrappers are smart contracts that "wrap" the settlement process, executing custom logic surrounding the settlement contract. When a solver executes a settlement that includes a wrapper, they call the wrapper contract instead of the settlement contract directly. The wrapper calls the settlement contract on behalf of the solver.
+
+This mechanism extends CoW Protocol's functionality in a modular way, allowing new features and integrations to be added without modifying the core settlement contract or requiring any changes to solver implementations.
+
+## Use Cases
+
+As we begin to roll out wrappers more widely across our infrastructure, we look forward to building out new advanced trading and DeFi integrations:
```
pretf00d: Re-work sentence. 

As we begin to roll out wrappers more widely across our infrastructure, below is an overview of some of the ways Wrappers can be used.

pretf00d (changes requested):
one minor tweak

docs/cow-protocol/concepts/order-types/wrappers.md
kaze-cow: e2a761f

docs/cow-protocol/integrate/wrappers.mdx
```diff
@@ -0,0 +1,365 @@
...
+
+This section covers building new wrapper contracts from scratch.
+
+### Overview
+
+To build a wrapper, you will:
+
+1. Inherit from the `CowWrapper` abstract contract
+2. Implement `_wrap()` with your pre/post-settlement logic
+3. Implement `parseWrapperData()` for input validation
```
fedgiac:
```suggestion
3. Implement `validateWrapperData()` for input validation
```

docs/cow-protocol/integrate/wrappers.mdx
```diff
@@ -0,0 +1,365 @@
...
+Before submitting an order, you can confirm the wrapper encoding is valid using [`CowWrapperHelper`](https://gist.github.com/kaze-cow/8c8a49f3305519948b1ec868c7efa5b9):
+
+```typescript
+// On-chain validation (via Ethers/Viem)
+const helper = new ethers.Contract(HELPER_ADDRESS, HELPER_ABI, provider)
+
+const isValid = await helper.verifyAndBuildWrapperData(
+    ["0x1234...", "0x5678..."],  // Wrapper addresses
+    ["0xabcd...", "0xef01..."]   // Wrapper data
+)
```
fedgiac: Inconsistent with the actual function:

```
    function verifyAndBuildWrapperData(WrapperCall[] memory wrapperCalls)
```

docs/cow-protocol/integrate/wrappers.mdx
```diff
@@ -0,0 +1,365 @@
...
+// Returns encoded wrapperData if valid, reverts if invalid
+```
+
+This checks:
+- All wrappers are allowlisted
+- All wrappers use the same settlement contract
```
fedgiac: This isn't the case.

```suggestion
```

docs/cow-protocol/integrate/wrappers.mdx
```diff
@@ -0,0 +1,365 @@
...
+- Each wrapper can parse its data successfully
+
+Note: an official deployment of the `CowWrapperHelper` contract does not currently exist. If you would like to use this contract, please deploy it yourself using the [code in the gist](https://gist.github.com/kaze-cow/8c8a49f3305519948b1ec868c7efa5b9).
```
fedgiac: Not needed for the docs PR, but why not? We could easily deploy that.

docs/cow-protocol/integrate/wrappers.mdx
```diff
@@ -0,0 +1,365 @@
...
+```
+
+**Fields:**
+- **`target`**: Address of the wrapper contract (treat as opaque - don't parse)
+- **`data`**: Wrapper-specific data (treat as opaque - don't parse)
+- **`isOmittable`**: Critical flag for solver requirements (see below)
+
+:::warning
+Wrapper addresses and data should be treated as opaque values. Do not attempt to parse or validate them - simply pass them through in your encoding.
+:::
```
fedgiac: Why are we so strict about this?
There is no real drawback in actually trying to parse if a solver wants to. They might have a good reason for it.
They kind of have to anyway, for example if `target` isn't an address, what should they do? What if the calldata is literally `0xinvalid`?
I think we should just remove all of this and just link to the section `### AppData Structure`.

docs/cow-protocol/integrate/wrappers.mdx
```diff
@@ -0,0 +1,365 @@
...
+
+#### 1. Execute Non-Omittable Wrappers
+
+Orders with `"isOmittable": false` **MUST** be executed with the specified wrapper. You may be slashed for not doing so.
+
+If `"isOmittable": true`, you MAY skip the wrapper if you find a better solution without it.
+
+#### 2. Verify Wrapper Authentication
+
+All approved wrappers will be approved by the DAO and registered in [`GPv2AllowlistAuthenticator`](../reference/contracts/core/allowlist.md). It is recommended to verify wrappers are allowlisted before including them in settlements.
```
fedgiac: Here instead it would be good to be strict. We're letting users tell the solver what to call and solvers should be aware of this.

```suggestion
All approved wrappers will be approved by the DAO and registered in [`GPv2AllowlistAuthenticator`](../reference/contracts/core/allowlist.md).

:::warning
The wrapper will be _directly_ called from the solver address and will mediate the settlement.
This value is provided _directly_ by the user and it may be malicious.
:::
```

docs/cow-protocol/integrate/wrappers.mdx
```diff
@@ -0,0 +1,365 @@
...
+
```
fedgiac, line 182: I think there should be an extra step here: decide whether to include the order in the settlement or not.
Since wrappers are likely to have complex behaviors, they would have to assess the revert risk based on the concrete wrapper implementation and may decide to skip it or adjust the scoring of the settlement accordingly.

docs/cow-protocol/integrate/wrappers.mdx
```diff
@@ -0,0 +1,365 @@
...
+]);
+// This is your settleData - unchanged from normal settlement
+```
+
+##### 3. Encode wrapperData Parameter
+
+The `wrapperData` combines individual wrapper data with wrapper addresses:
+
+```typescript
+function encodeWrapperData(wrappers: WrapperCall[]): string {
```
fedgiac: Can we link to a reference here instead of adding code?
Something I'm worried about is that there's no input validation whatsoever (we actually say "WARNING: don't validate"), then this would introduce super weird behaviors like a user injecting data in the `target` field and causing inconsistencies like being able to inject other wrappers or calls that aren't part of the expected execution.
We're also making a lot of assumptions, like the data being `0x` prefixed. Is it ok if the solvers skips the first byte provided by the user?
I'd say the user data should actually be validated at some point: it should conform to the app data schema, `target` should be an address, and `data` should be correctly encoded bytes, all `0x`-prefixed.

docs/cow-protocol/integrate/wrappers.mdx
```diff
@@ -0,0 +1,365 @@
...
+
+For 3 wrappers:
+[length₁(2B)][data₁][address₂(20B)][length₂(2B)][data₂][address₃(20B)][length₃(2B)][data₃]
+```
+
+##### 4. Call wrappedSettle
+
+```typescript
+const tx = {
+    to: wrappers[0].target,
```
fedgiac, line 239: It's nice to point out that the first target is missing in the encoding because it appears here.

docs/cow-protocol/integrate/wrappers.mdx
```diff
@@ -0,0 +1,365 @@
...
+bytes memory wrapperData = CowWrapperHelper(HELPER).verifyAndBuildWrapperData(
+    wrapperAddresses,
+    wrapperDatas
+);
+```
+
+In addition to producing the correct encoded data, the helper also validates:
+- All wrappers are allowlisted
+- All wrappers use the same settlement contract
+- Each wrapper can parse its data
```
fedgiac: I've seen this appearing before, possibly verbatim. We should just add a link, in any case the comments from before apply.

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
```diff
@@ -0,0 +1,630 @@
...
+## Implementation
+
+Developers looking to integrate using wrappers should copy this all-in-one solidity file into their project as vendored dependency.
+
+It provides a few base contracts which can serve as the foundation for integration:
+
+- **`ICowWrapper`**: Core interface all wrappers must implement
+- **`CowWrapper`**: Abstract base contract providing security and utilities that all wrappers should use
+
+```solidity
```
fedgiac: Still suggesting to not copy the code here but just have a link.
- I don't believe we'll keep this up-to-date once a change is made, we're going to forget this dependency exists.
- If I were writing a wrapper, I'd like to have a more authoritative/better tracked source for this critical part of the wrapper and not just "I copied this from the textual docs."

Previous discussion [here](https://github.com/cowprotocol/docs/pull/595#discussion_r2854599329).

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
```diff
@@ -0,0 +1,630 @@
...
+        }
+    }
+}
+```
+
+### Quick Start Example
+
+Here's a minimal wrapper implementation to use as a starting point:
+
+```solidity
```
fedgiac: I don't like to see large swathes of code in the docs.
I'd just like to make the code in the docs as minimal as possible because the docs are not a dev environment. This is better for maintainability, reproducibility, testing. The problems for a reader:
- Is this code up to date? If not, when was this working? What tooling did it use, like, Solidity version and Foundry version?
- I want to make a change for clarity. Does my code still work after that?
- I'm reviewing this code. Am I the one responsible for making sure the code works?

We can create a new repo, "generalized wrapper examples" that we can properly test and keep updated.

Case in point, it doesn't compile.

```
$ forge build
[⠊] Compiling...
[⠔] Compiling 5 files with Solc 0.8.30
[⠒] Solc 0.8.30 finished in 376.14ms
Error: Compiler run failed:
Error (3656): Contract "MyWrapper" should be marked as abstract.
 --> src/MyWrapper.sol:6:1:
  |
6 | contract MyWrapper is CowWrapper {
  | ^ (Relevant source part starts here and spans across multiple lines).
Note: Missing implementation: 
  --> src/CowWrapper.sol:80:5:
   |
80 |     function name() external view returns (string memory);
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Error (4649): Invalid type for argument in modifier invocation. Invalid implicit conversion from address to contract ICowSettlement requested.
 --> src/MyWrapper.sol:7:48:
  |
7 |     constructor(address settlement) CowWrapper(settlement) {}
  |                                                ^^^^^^^^^^
```

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
fedgiac: Moving the discussion [here](https://github.com/cowprotocol/docs/pull/595#discussion_r2918151340) so I don't need to unfurl conversations on GitHub to see open comments.

docs/cow-protocol/integrate/wrappers.mdx
kaze-cow: I suppose this could be reworded, but one common confusion for solvers is that they think they need to understand what is going on and how they should set up the actual solution. There have been multiple solvers during implementation that have asked me how they are supposed to parse/understand this data, and I have had to clarify that it is intentionally designed to be *not necessary* to do this. So this comment is designed to explicitly disperse those types of confusions. From a security perspective, yes, it seems like there is some use in that, but if solvers have to be so defensive about the use of wrappers, we have an unsurmountable problem in the design of wrappers in and of themselves. 

If you ask me, I think it would be more effective for a solver to **simulate the call** that would be executed and parse the simulation to see how it behaves (ex. what addresses it touches, funds transfers, etc.). Attempting to parse the wrapper data is kind of a losing game because we want to be able to add more new wrappers at any time and expecting solvers to maintain their own allowlist of solvers is going to make it much harder for us to introduce new wrappers in the future.

Also regarding `target` not being an address, [the `appData` schema](https://github.com/cowprotocol/cow-sdk/blob/main/packages/app-data/src/schemas/wrappers/v0.2.0.json#L18) literally constrains it to be a [hex with 40 characters](https://github.com/cowprotocol/cow-sdk/blob/main/packages/app-data/src/schemas/definitions.json#L16) (ignoring checksumming). [Same happens in the autopilot and api](https://github.com/cowprotocol/services/blob/main/crates/app-data/src/app_data.rs#L71) when they receive app data.

docs/cow-protocol/integrate/wrappers.mdx
kaze-cow: Do we have any specific advice on how to actually do that? I don't, other than simulating the order (which we already do in the driver and sometimes autopilot...). Seems like something we have to discuss. See previous comment about adding new wrappers and lack of solver involvement.

docs/cow-protocol/integrate/wrappers.mdx
kaze-cow: > We're also making a lot of assumptions, like the data being 0x prefixed. Is it ok if the solvers skips the first byte provided by the user?

By the time the solver receives the data it will have been parsed into bytes by the autopilot engine as far as I can see. I agree it would probably be safer to use some sort of a "hex concat" function, but that requires an external library, idk if we have established use of such a thing yet. (afaict javascript does not have a suitable hex concatenation function; it always assumes the 0x doesn't exist)

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
kaze-cow: alright since no great solution exists right now to have it embedded its just going to be linked as suggested.

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
kaze-cow: This was never intended to be a fully fledged example of a real wrapper, just a segment of code that has the basic pieces to hit the ground running (and apparently missing the `name` function, but resolvable). As an alternative, do you think it would be better to simply document the individual pieces that are required (ex. `_wrap`, `validateWrapperData`, `name`)? I suppose we kind of already go over that in the lower section "virtual functions for integrators"

docs/cow-protocol/integrate/wrappers.mdx
kaze-cow: [8d1e9a0](https://github.com/cowprotocol/docs/pull/595/commits/8d1e9a017526aa392008d9b4d26ca97da0e59b45)

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
kaze-cow: [8d1e9a0](https://github.com/cowprotocol/docs/pull/595/commits/8d1e9a017526aa392008d9b4d26ca97da0e59b45)

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
kaze-cow: removed this section because its kind of a duplicate of existing sections which explain the format pretty well.

kaze-cow: 42ac22b

docs/cow-protocol/integrate/wrappers.mdx
kaze-cow: pointed out 42ac22b

docs/cow-protocol/integrate/wrappers.mdx
kaze-cow: linked to elsewhere, tis well docced in this repo already 42ac22b

docs/cow-protocol/integrate/wrappers.mdx
kaze-cow: Added additional docs here advising to simulate the transaction and perform any checks that may be necessary, especially confirming the transaction reverts or not 42ac22b

docs/cow-protocol/integrate/wrappers.mdx
kaze-cow: after private discussion we decided to indeed reword this and to make sure we express that outside of basic validation, deeply understanding the data in the wrapper is not necessary, but we recommend solvers take measures to ensure security 42ac22b

docs/cow-protocol/integrate/wrappers.mdx
kaze-cow: I am adding the deployment of this to the Cannon package and put the addresses that CREATE2 generates in :+1: 42ac22b  https://github.com/cowprotocol/euler-integration-contracts/pull/25

docs/cow-protocol/integrate/wrappers.mdx
kaze-cow: consistantified 42ac22b

kaze-cow:
benyond just your general comments @fedgiac , I explicitly tried to remove anything repetative from the docs, explicitly prioritizing links to maintainable code on github or other pages if the doc is already there, throughout all pages. LMK if I missed anything!

docs/cow-protocol/integrate/wrappers.mdx
```diff
@@ -0,0 +1,322 @@
...
+      "isOmittable": false
+    }
+  ]
+}
+```
+
+**Fields:**
+- **`target`**: Address of the wrapper contract (treat as opaque - don't parse)
+- **`data`**: Wrapper-specific data (treat as opaque - don't parse)
+- **`isOmittable`**: Whether the `target` and `data` must be executed unmodified as part of the order execution
```
fedgiac: ([Continuing](https://github.com/cowprotocol/docs/pull/595#discussion_r2922649447).)
I see "parsing" as the process of interpreting the opaque string to the expected format (i.e., one Ethereum address and one bytestring). Not parsing means that solvers must literally accept `"target": "invalid address"`. The cost of doing this is negligible.

> Also regarding target not being an address, [the appData schema](https://github.com/cowprotocol/cow-sdk/blob/main/packages/app-data/src/schemas/wrappers/v0.2.0.json#L18) literally constrains it to be a [hex with 40 characters](https://github.com/cowprotocol/cow-sdk/blob/main/packages/app-data/src/schemas/definitions.json#L16) (ignoring checksumming). [Same happens in the autopilot and api](https://github.com/cowprotocol/services/blob/main/crates/app-data/src/app_data.rs#L71) when they receive app data.

Users can sign an order with nonsense app data (at least, this was the case last time I checked, through the ETH flow). Anyway, even if the orderbook rejects it, that's still a valid order.
We don't provide any guarantee that the app data is valid, since it's user controlled. We don't want to have the orderbook guarantee the validity of the app data. **The orderbook is not a gatekeeper for the protocol.** We don't want it to become one. Points of centralization should be avoided and this assumption would be a small step towards centralization.

Also, we gain really nothing out of this. What's the point of saving an address check and a bytes decoding for the solver?

What I see being the point here is that we don't want solvers to "decode" the data, the mechanism is the same for all wrappers.

```suggestion
- **`target`**: Address of the wrapper contract
- **`data`**: Wrapper-specific data
- **`isOmittable`**: Whether the `target` and `data` must be executed unmodified as part of the order execution

`target` may be any address and `data` is arbitrary bytes that will be passed to the target. There is generally no need for decoding the wrapper data since the execution process uses the same opaque format on each wrapper.
```

docs/cow-protocol/integrate/wrappers.mdx
```diff
@@ -0,0 +1,322 @@
...
+```
+
+This checks:
+- All wrappers are allowlisted
+- Each wrapper can parse its data successfully
+
+The CoW Wrapper Helpers contract is deployed via `CREATE2`, and is being rolled out to the following addresses on each network supported by CoW:
+* Production: `0x8C8f88F3081A7d54fCbC2Bd7C6Cf5E9Fc26e30E9`
+* Staging: `0xfd1a35C8E7AEC58a4bB6AeDca0B2Bf9c9Be9E07F`
+
```
fedgiac: Can you move this to [this page](https://docs.cow.fi/cow-protocol/reference/contracts/periphery) and add a link here for consistency?
Eventually I hope we change how we generate that page thanks to the new deployment repo!

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
```diff
@@ -0,0 +1,305 @@
...
+    ) external view returns (bytes memory chainedWrapperData);
+}
+```
+
+## Implementation Requirements for Integrators
+
+### Security Requirements
+
+:::warning Critical Requirements
+
```
fedgiac: Tag is never closed.

```suggestion
```

docs/cow-protocol/integrate/wrappers.mdx
```diff
@@ -0,0 +1,322 @@
...
+To build a wrapper, you will:
+
+1. Inherit from the `CowWrapper` abstract contract
+2. Implement `_wrap()` with your pre/post-settlement logic
+3. Implement `validateWrapperData()` for input validation
+4. Test thoroughly on testnets
+5. Review the [Implementation Requirements](../reference/contracts/periphery/wrapper.mdx#implementation-requirements-for-integrators) to ensure your wrapper meets the acceptance requirements
+6. Submit for allowlist approval by the CoW Protocol team
+
```
fedgiac, line 26: Let's stress that it's difficult to build one. We want to prevent people to expend effort for nothing and we want to avoid spam submissions.

```suggestion

Building a wrapper is a significant undertaking.
An insecure wrapper would put user and protocol funds at risk.
Make sure you understand CoW Protocol well before starting.
A contract audit by a reputable firm is mandatory before a wrapper is considered for approval.
```

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
```diff
@@ -0,0 +1,305 @@
...
+- Calls the next wrapper via `wrappedSettle()`, or the settlement contract via `settle()` if no wrappers remain
+
+## Calldata Encoding Specification
+
+The `chainedWrapperData`  parameter of `wrappedSettle(settleData, chainedWrapperData)` is specially encoded to minimize bit shuffling overhead:
+
+```
+┌──────┬──────────┬──────────┬──────┬──────────┐
+│ Len₁ │  Data₁   │  Addr₂   │ Len₂ │  Data₂   │
+│(2 B) │ (wrap1)  │ (20 B)   │(2 B) │ (wrap2)  │
```
fedgiac: For clarity, since `wrap1` and `wrap2` come out of nowhere.

```suggestion
│ Len₁ │  Data₁   │  Addr₂   │ Len₂ │  Data₂   │
│(2 B) │  (Len₁)  │ (20 B)   │(2 B) │  (Len₂)  │
```

docs/cow-protocol/reference/contracts/periphery/wrapper.mdx
```diff
@@ -0,0 +1,305 @@
...
+    function validateWrapperData(bytes calldata wrapperData)
+        external
+        view
+        override
+    {
+        // No validation needed
+    }
+}
+```
+
```
fedgiac: Similar comment to [this](https://github.com/cowprotocol/docs/pull/595#discussion_r2918177289), we already link to this contract below.


```suggestion
```

fedgiac (commented):
Final comments.

docs/cow-protocol/integrate/wrappers.mdx
fedgiac: [Continuation](https://github.com/cowprotocol/docs/pull/595#discussion_r2930271284).

docs/cow-protocol/integrate/wrappers.mdx
fedgiac: > which we already do in the driver and sometimes autopilot

The solver doesn't have to use the official driver. The driver does some checks, but solvers should know which of them are needed. A solver may be using the official driver, but it doesn't have to and the docs should reflect that.
For the autopilot, "sometimes" isn't enough, and even if it was "always," that's still not ideal because we don't want to add more responsibilities to the autopilot. We shouldn't use the autopilot as a gatekeeper (similar discussion [here](https://github.com/cowprotocol/docs/pull/595#discussion_r2930271284)).

fedgiac (commented):
Looks good to me. Not approving because the build failed, but once that's sorted out it can be merged.

fedgiac (approved):

pretf00d (approved):

PR #650, opened 2026-08-03
Improve docs for solver API: limit amounts vs full amounts

tamir-cow:
Clarify that sellAmount/buyAmount reflect an order's remaining (unfilled) amount, while fullSellAmount/fullBuyAmount are the original signed amounts that stay constant across auctions.

## Summary by CodeRabbit

* **Documentation**
  * Clarified the distinction between remaining executable order amounts and original signed amounts.
  * Documented the new `fullSellAmount` and `fullBuyAmount` fields in the order schema.
  * Updated examples and explanations to reflect partial order execution and volume-based fees.


docs/cow-protocol/reference/core/auctions/schema.md
```diff
@@ -76,8 +76,10 @@ This key maps to a list containing the set of orders in the batch. Each entry in
 - `uid`: this is the unique identifier of the order.
 - `sellToken`: a string denoting the address of the sell token.
 - `buyToken`: a string denoting the address of the buy token.
-- `sellAmount`: a stringified integer denoting the limit amount that is being sold, measured in terms of the smallest denomination of the sell token.
-- `buyAmount`: a stringified integer denoting the limit amount that is being bought. Similar to the `sellAmount`, it is measured in terms of the smallest denomination of the buy token.
+- `sellAmount`: a stringified integer denoting the limit amount that is being sold, measured in terms of the smallest denomination of the sell token. For a partially fillable order that has already been filled in a previous auction, this is the *remaining* amount that is still available to be matched (i.e., it already accounts for the `executed` amount), and is the value solvers should use when computing a solution.
```
fhenneke: There are two reasons for `sellAmount` and `fullSellAmount` being different.
* For partially fillable orders, `sellAmount` takes into account that some part of the sell amount of the can could already have been executed. That part is explained in the new text.
* For all orders with volume fee (which are essentially all orders), `sellAmount` is modified such that after charging all volume fees one ends up with `fullSellAmount`. This Is done so that solvers can just use `sellAmount` in their routing and ignore all fees, while the driver can (in most cases) charge protocol/partner fees without violating the limit amount `fullSellAmount`. This aspect is not reflected in this PR.

I think the second point should also be addressed as it is relevant for almost every order.

fhenneke (changes requested):

docs/cow-protocol/reference/core/auctions/schema.md
```diff
@@ -107,14 +109,16 @@ An example Fill-or-Kill user limit buy order that sells 1000 [COW](https://ether
...
+The above entry should be interpreted as follows. It is a Fill-or-Kill order since the flag `partiallyFillable` is set to `false`. Moreover, it is a sell order since its `kind` is set to `sell`. Finally, this is a `limit` order, meaning that it has a zero-signed fee, which implies that the solver is free to choose an appropriate fee to cover its execution cost. This means that, if executed, the user will send a total of 1000000000000000000000 COW atoms to the settlement contract and, no matter how much fee the solver will charge, the user is guaranteed to receive at least 284138335 USDC atoms. Here, `sellAmount`/`buyAmount` equal `fullSellAmount`/`fullBuyAmount` because the order has not been partially filled and has no `protocolFees` applied. In general they can differ: partial fills reduce `sellAmount`/`buyAmount` to the remaining unfilled amount, and volume-based fees scale them down further, so the driver can charge those fees afterwards without violating the user's original limit.
```
fhenneke: This case of `buyAmount = fullBuyAmount` does not happen in practice on most chains. This example should cover the usual case, I think, of a 2bps volume fee. The given signed amount of `284138335` would be consistent with a `buyAmount` of `284138335 / (1 - 0.0002) = 284195174.03...` rounded to `284195174`.

docs/cow-protocol/reference/core/auctions/schema.md
```diff
@@ -106,15 +108,23 @@ An example Fill-or-Kill user limit buy order that sells 1000 [COW](https://ether
...
     "kind": "sell",
     "partiallyFillable": false,
-    "class": "limit"
+    "class": "limit",
+    "protocolFees": [
+        {
+            "kind": "volume",
+            "factor": 0.0002
+        }
+    ]
```
fhenneke: I would revert this change. Fee policies are not always forwarded to solvers. (This is also not documented cleanly anywhere, I had to ask Claude to give some pointers on this.)
As an example, [this auction instance](https://solver-instances.s3.eu-central-1.amazonaws.com/prod/mainnet/auction/13555388.json) does have the modified amounts but not any fee policies.

docs/cow-protocol/reference/core/auctions/schema.md
```diff
@@ -106,15 +108,23 @@ An example Fill-or-Kill user limit buy order that sells 1000 [COW](https://ether
...
...st. This means that, if executed, the user will send a total of 1000000000000000000000 COW atoms to the settlement contract, and the user has signed for a minimum of 284138335 USDC atoms (`fullBuyAmount`). The order also carries a `protocolFees` entry for a 2 bps volume fee, which the driver will deduct from the buy side once the trade settles. Since the order has not been partially filled, `sellAmount` still equals `fullSellAmount`, but `buyAmount` is scaled up from `fullBuyAmount` to net out that fee: `buyAmount` = `fullBuyAmount` / (1 - 0.0002) = 284138335 / 0.9998 ≈ 284195174.03, rounded down to 284195174. Solvers can then route on `sellAmount`/`buyAmount` while ignoring fees entirely, and once the driver charges its 2 bps cut of the 284195174 USDC bought (≈56839 USDC atoms), the user is left with exactly the signed minimum of 284138335 USDC atoms — so the original limit is never violated. This is the usual case in practice: on most chains essentially every order carries some volume-based protocol or partner fee, so `buyAmount`/`sellAmount` differ from `fullBuyAmount`/`fullSellAmount` even without any partial fill; the two only coincide when there is no such fee, which is rare.
```
fhenneke: Probably rephrase this then from

> The order also carries a `protocolFees` entry for a 2 bps volume fee, which the driver will deduct from the buy side once the trade settles.

to

> The order also carries a 2 bps volume fee, which the driver will deduct from the buy side once the trade settles.

fhenneke (changes requested):
I would remove the info on the fee policy entry. Otherwise looks good now.

fhenneke (approved):

raq0x (approved):

PR #655, opened 2026-08-10
CIP: Penalty Cap Redesign

fhenneke:
Documentation change reflecting the [penalty cap redesign CIP draft](https://forum.cow.fi/t/cip-draft-penalty-cap-redesign/3520): the chain-specific absolute penalty cap is replaced by a per-order penalty cap.

This PR would be merged, after filling in the `CIP-XX` placeholders, if the CIP were to pass.

## Summary by CodeRabbit

* **Documentation**
  * Updated solver rewards documentation to include CIP-87..
  * Clarified performance reward caps, including chain-specific upper limits and lower limits based on unsettled orders.
  * Documented per-order penalty caps using volume, a global 20 USD limit, fill fraction, and chain/token-pair basis points.
  * Expanded solver strategy guidance to explain how penalties are bounded.

PR #657, opened 2026-08-25
added optimal bidding strategy page to docs tutorials

Architsharma7:
# Description

Adds a solver-facing guide to the docs tutorials covering the recommended baseline bidding strategy for CoW Protocol's combinatorial auction

ref: [Notion](https://app.notion.com/p/cownation/Optimal-Bidding-Strategy-Docs-36b8da5f04ca80c39b68d9c724391d5d?source=copy_link)


Architsharma7:

fhenneke (commented):
One thing I would change in the tutorial is to be more detailed in the explanation of the initial formula for bidding, and way less detailed in explaining how some other aspects of the competition change behavior. The latter seems somewhat unnecessary for a tutorial, while also inviting lots of subtle errors. I would not be confident in proposing how consistency rewards change strategy, or what solvers should do in the face of upper reward caps. Instead, I would frame it as "these other things have impact on bidding as well, feel free to experiment with changing your strategy away from the baseline".

Of the extensions, only slippage seems to fall into the scope of this tutorial. The worked example should be extended to that.

I think that the approach of the knowledge sharing session worked well. I.e. going from routing, to value, to risk adjusted bid, to slippage tolerance.

I added a Claude review below which touches on a few valid points and lots of nitpicks.

----------

Review by Claude, run by @fhenneke.

## Reward cap, consistency rewards, quote rewards

These three sections restate mechanism details at a level of precision that introduces errors:

- `c_u` is treated as an exogenous chain parameter. It is `β · protocolFee` (excluding partner fees) over the solver's solutions settled in that auction, so it is not independent of the solution.
- The `(1−α)·c_l` refinement omits that an unset also lowers the settlement success rate, which multiplies the consistency metric. The page therefore understates the cost of an unset.
- The quote-reward condition is not `s ≥ q`. Per CIP-72 the requirement is that the quoting solver proposed an execution of that order at least as good as the quote and that it was not removed by fairness filtering, plus three further conditions (fill-or-kill market order, verified quote, order executed). A score and a quoted amount are different quantities.

Suggest replacing the three sections with one stating that the reward cap, consistency rewards and quote rewards also affect the profitability of a bid, linking to the [rewards reference](https://docs.cow.fi/cow-protocol/reference/core/auctions/rewards).

The same applies to the Dune table, which will go stale. The text says ~2,000 auctions, the table says 1,464.

## Corrections

1. Unset penalty. The page states `min(reference score, c_l)`. Per the rewards reference it is `min(referenceScore_i − score of the other winners, c_l)`, which collapses to `min(referenceScore, c_l)` only when the solver is the sole winner of the auction, not when it has a single winning solution.
2. Score composition. The page has `s(x) = user surplus + protocol fee + partner fee`. the-problem defines `score(o) = (U(o) + f(o))·p(b)` with `f` the protocol fee, and the rewards page says "protocol fee (excluding partner fees)" for the cap. Whether partner fees enter the score needs confirming; if they do, the-problem should state it as well.
3. Positive score is a validity condition. `s_base` can be zero or negative for low-value, low-`p` candidates, which conflicts with "submit every viable candidate".
4. Slippage tolerance. The derivation substitutes the reported score for the reference score without saying so, and drops the reward term, which is only valid under that substitution.
5. Worked example. Solution C is `0.42 − (0.06/0.94)·0.01 = 0.419362`, so `0.4194`, not `0.4193`.
6. `c_l = 0.01 ETH` is hardcoded. It is chain-specific and may change (#655). Link the chain table instead.

## Duplication

`rewards.md` § Solver's strategy already covers truthful bidding, the cap's first-price logic and fairness-filter strategy. The third paragraph of "The auction setting" is close to verbatim from it.

## Framing

The TL;DR ("the score should be neither inflated ... nor reduced") reads as prescriptive. Consider stating once that this is a starting point rather than a competition rule, and that the optimal strategy for this mechanism is not known.

## Nits

- No internal links. `rewards`, `the-problem`, `competition-rules#off-chain-protocol` and `accounting` are candidates.
- No `sidebar_position` in the frontmatter; every sibling page has one, so this page sorts last.
- `$min(...)$` renders as italic *min*; use `$\min(...)$`.
- "If on-chain slippage would exceeds $-\gamma$": grammar, and the sign conflicts with "maximum negative slippage".
- "is sum of three components" → "is the sum of"; "incurs the a penalty"; "an heuristic" → "a heuristic"; "i.e," → "i.e.,".
- Overview: "estimate the value it creates, the probability it settles successfully" is missing a conjunction.
- Checklist item 1 repeats the section's lead-in ("For each candidate solution $x$").

docs/cow-protocol/tutorials/solvers/optimal_bidding.md
```diff
@@ -0,0 +1,178 @@
...
+
+:::info
+**TL;DR: How to bid?**
+
+For each solution a solver can settle, the recommended bid is a score equal to the value the solution is expected to deliver, adjusted downward for the probability that the settlement does not land on chain. The score should be neither inflated to win more auctions nor reduced to retain more value: under the standard mechanism, the reported score determines only whether the solution wins, not the solver's payoff conditional on winning. The solver should also set a threshold on acceptable negative slippage so that a solution is settled only when its realised value remains non-negative, and should submit every solution it can profitably execute.
+:::
+
+## Overview
+
+For each candidate solution $x$ the solver can submit, estimate the value it creates, the probability it settles successfully. Use the risk-adjusted score as the baseline bid. This is a dominant strategy in a simplified mechanism explained below and a recommended baseline in production. Deviations from the baseline require data about competitor scores, reward-cap binding, or fairness-filter effects.
```
pretf00d: estimate the value it creates, the probability it settles successfully. --> estimate the value it creates, and the probability it settles successfully.

docs/cow-protocol/tutorials/solvers/optimal_bidding.md
```diff
@@ -0,0 +1,178 @@
...
+## Score and value
+
+Let $x$ denote a candidate solution: the set of orders to be executed, their effective executed amounts, and the routing/liquidity sources used to execute them.
+
+There are two main ways to assign value to a solution. Conflating them is the most common source of confusion.
+
+- $S(x)$: the solver’s estimate of total score-relevant value created by solution $x$, net of execution costs (gas, AMM fees, slippage against liquidity sources used in the route).
+- $s(x)$: the score reported to the protocol for solution $x$.
+
+The reported score induced by the solution $x$ is sum of three components:
```
pretf00d: is sum of three components --> is the sum of three components

docs/cow-protocol/tutorials/solvers/optimal_bidding.md
```diff
@@ -0,0 +1,178 @@
...
+
+$$
+s(x) = (\text{user surplus}) + (\text{protocol fee}) + (\text{partner fee})
+$$
+
+The user receives only the user surplus component. The remaining two are value generated by the trade that the protocol and any integrating partner collect.
+
+The solver retains the difference $S(x) - s(x)$, typically held as buffers in the settlement contract and reconciled weekly (a solver could also instead transfer the amount to itself immediately in the settled token). The guidance below assumes scores correspond to actual economic value in the unit the protocol uses for scoring.
```
pretf00d: could also instead transfer --> could also transfer

docs/cow-protocol/tutorials/solvers/optimal_bidding.md
```diff
@@ -0,0 +1,178 @@
...
+
+Solvers can choose not to settle a selected solution, for example, if the desired routing would return significantly fewer tokens than expected. One way to implement this is to set a slippage tolerance $\gamma$: the maximum negative slippage the solver is willing to absorb between bidding and settlement. If on-chain slippage would exceeds $-\gamma$, the settlement does not happen.
+
+A conservative baseline per submitted solution:
+
+$$
+\gamma(x) = S(x)-s(x)+\min(c_l,s(x))
+$$
+
+The solver compares the payoff from settling through adverse slippage with the realised payoff from unsetting. Settling yields approximately $S(x)−s(x)− \gamma$. Unsetting instead incurs the a penalty $min(s(x),c_l)$. The slippage tolerance is therefore the point at which the solver is indifferent between these two outcomes.
```
pretf00d: Unsetting instead incurs the a penalty --> Unsetting instead incurs a penalty

docs/cow-protocol/tutorials/solvers/optimal_bidding.md
```diff
@@ -0,0 +1,178 @@
...
+
+For small solutions where $s(x)\leq c_l$, this reduces to $\gamma(x)\approx S(x)$. For larger solutions, it becomes $\gamma(x)\approx S(x)-s(x)+c_l$, meaning the solver tolerates less negative slippage because the cost of unsetting is capped.
+
+## Fairness filtering
+
+A batched solution is excluded by the fairness filter if it under-delivers on any directed token pair relative to the per-pair reference outcome (constructed from the best individual-pair solutions across all solvers).
+
+Two operational implications:
+
+**Submit individual-pair fallbacks.** If a solver has a batched solution covering multiple pairs, also submit the underlying individual-pair solutions. A high-scoring batched solution that fails fairness is worth zero, the individual-pair fallbacks ensure the solver still wins the pairs it can.
```
pretf00d: is worth zero, the individual-pair fallbacks ensure the solver still wins the pairs it can. --> is worth zero. The individual-pair fallbacks ensure the solver still wins the pairs it can.

docs/cow-protocol/tutorials/solvers/optimal_bidding.md
```diff
@@ -0,0 +1,178 @@
...
+
+**Verify batched solutions pass the filter.** Compute the per-pair reference outcome from the best individual-pair solutions in the auction and check that the batched solution delivers at least that much on every directed pair it touches.
+
+Individual-vs-batch submission can also be strategic: strong individual-pair bids raise the fairness benchmark and can exclude competitors’ batched solutions, and vice versa. The baseline strategy in this guide does not rely on strategically changing scores to affect the fairness filter, and we do not recommend solvers use the fairness filter as a strategic target. Instead, solvers should compute honest individual-pair solutions, submit batched solutions only when they create additional value, and verify that those batched solutions pass the fairness filter.
```
pretf00d: we do not recommend solvers use the fairness filter" → drops into first-person "we" for one sentence; rest of the doc is third-person/imperative. Suggest you pick one voice

docs/cow-protocol/tutorials/solvers/optimal_bidding.md
```diff
@@ -0,0 +1,178 @@
...
+
+## Consistency rewards
+
+[CIP-85](https://snapshot.box/#/s:cow.eth/proposal/0xb488c343df3ba5f3857a4c7a920a74e18c13a2cdce99d27af34216803da6abff) distributes a weekly consistency budget across solvers based on participation. Penalties paid on unsets feed this budget, so a solver receives back a fraction $\alpha$: their share of the budget at week-end, which is non-trivial to estimate.
+The effective per-revert protocol cost is therefore $(1-\alpha)c_l$, not $c_l$. The baseline formula above ignores this. Solvers with a stable consistency share can refine the cap-binding term by replacing $c_l$ with $(1-\alpha)c_l$ which increases optimal bids when the penalty cap binds. The exact $\alpha$ depends on aggregate weekly metrics that themselves depend on other solvers' behaviour, so the refinement only pays off for solvers who can estimate their share reliably and hence is used as an heuristic.
```
pretf00d: an heuristic --> a heuristic

docs/cow-protocol/tutorials/solvers/optimal_bidding.md
```diff
@@ -0,0 +1,178 @@
...
+
+This means quote rewards should be incorporated into the solver’s expected-value calculation rather than treated separately from bidding. If satisfying the quote-reward condition requires reporting a score above the baseline score, the solver should account for the additional probability of winning and the expected payoff of those extra wins. Whether this is profitable depends on the quote reward, the distribution of competing scores, and the solver’s settlement risk.
+
+The baseline score in this guide should therefore be read as the starting point before quote-reward incentives. Solvers participating in quote competition should incorporate the quote reward into their own expected-profit model.
+
+## Practical checklist
+
+For each candidate solution $x$:
+
+1. For each candidate solution $x$, estimate $S(x)$, $p(x)$.
```
pretf00d: duplicate of the above.

pretf00d (changes requested):

docs/cow-protocol/tutorials/solvers/optimal_bidding.md
```diff
@@ -0,0 +1,193 @@
...
+
+The intended reader is someone running or planning to run a CoW Protocol solver who wants a self-contained operational reference.
+
+:::info
+**TL;DR: How to bid?**
+
+For each solution a solver can settle, the recommended bid is a score equal to the value the solution is expected to deliver, adjusted downward for the probability that the settlement does not land on chain. The score should be neither inflated to win more auctions nor reduced to retain more value: under the standard mechanism, the reported score determines only whether the solution wins, not the solver's payoff conditional on winning. The solver should also set a threshold on acceptable negative slippage, so that a solution is settled unless the loss from slippage exceeds the cost of not settling, and should submit every solution with a positive baseline score.
```
fleupold: Have both prose and formula here (define variables in prose, then use in formula).

The last sentence reads weird (should be something like "a solution is settled only as long as the loss from slippage does not exceed the revert penalty). Also what is a baseline score and why can it be negative?

docs/cow-protocol/tutorials/solvers/optimal_bidding.md
```diff
@@ -0,0 +1,193 @@
...
+## The auction setting
+
+The protocol runs a **combinatorial auction** (see [competition rules](/cow-protocol/reference/core/auctions/competition-rules#off-chain-protocol)). In each auction, a solver can submit candidate **solutions**. A solution is a commitment to settle a specified set of orders together with a reported **score** $s$. A solution may cover:
+
+- A single directed token pair (handling all orders on that pair).
+- Multiple directed token pairs (a batched solution, useful when coincidence-of-wants (CoWs) between pairs allow for peer-to-peer trading or when shared gas improves the routing).
+
+The protocol filters batched solutions for fairness, then picks the combination of solutions across pairs and solvers that maximises total reported score.
+
+An operationally important constraint is that winner selection cannot choose two solutions that touch the same directed token pair, i.e., the same sell and buy tokens. If a solver wants to execute multiple orders on the same directed pair, those executions must be included in one combined solution. Multiple separate solutions on the same directed pair are not automatically aggregated by winner selection.
```
fleupold: How is this relevant for bidding?

docs/cow-protocol/tutorials/solvers/optimal_bidding.md
```diff
@@ -0,0 +1,193 @@
...
+
+For each winning solver, the protocol computes a performance reward by comparing the selected outcome with the reference outcome, i.e., the best outcome the protocol could have achieved without that solver. This reward is bounded by the lower penalty cap ($c_l$) and the upper reward cap ($c_u$) (see [rewards](/cow-protocol/reference/core/auctions/rewards#performance-rewards)).
+
+After winning, the solver is responsible for settling on chain. If the solver is the only winner of the auction and the settlement does not land before the deadline, the settlement is **unset**: the solver pays the protocol $\min(\text{reference score}, c_l)$. With multiple winning solutions, partial settlement failures require the corresponding more general accounting. This is why solvers should adjust reported scores for the risk of an unset.
```
fleupold: This reads weird. As if only multiple winning solutions require solvers to adjust reported scores. I don't think the differentiation between single winner and multiple winner is useful here (below for simplification sure).

docs/cow-protocol/tutorials/solvers/optimal_bidding.md
```diff
@@ -0,0 +1,193 @@
...
+
+This single-winner view provides a useful approximation for reasoning about how expected settlement risks should enter reported scores.
+
+## Score and value
+
+Let $x$ denote a candidate solution: the set of orders to be executed, their effective executed amounts, and the routing/liquidity sources used to execute them.
+
+There are two main ways to assign value to a solution. Conflating them is the most common source of confusion.
+
+- $S(x)$: the solver’s estimate of total score-relevant value created by solution $x$, net of execution costs (gas, AMM fees, slippage against liquidity sources used in the route).
```
fleupold:
```suggestion
- $S(x)$: the solver’s estimate of total value created by solution $x$, net of execution costs (gas, AMM fees, slippage against liquidity sources used in the route).
```

docs/cow-protocol/tutorials/solvers/optimal_bidding.md
```diff
@@ -0,0 +1,193 @@
...
+| Route output  |   17,110 |
+| User's limit  | − 17,000 |
+| Gas           |     − 10 |
+| **Value $S$** |  **100** |
+
+The solver now decides how to share these 100 USDC. Part goes to the user as surplus, part pays the order's fees, and the rest stays with the solver. The first two parts make up the score.
+
+If the order pays 10 USDC in fees and the solver promises the user 60 USDC of surplus, the score is $s = 70$ and the solver keeps $S - s = 30$.
+
+## Recommended baseline bid
```
fleupold:
```suggestion
## Recommended bid
```

Can we remove baseline from everywhere here? It's easy to conflate it with other concepts like the baseline solver (who doesn't use this bidding strategy)

docs/cow-protocol/tutorials/solvers/optimal_bidding.md
```diff
@@ -0,0 +1,193 @@
...
+
+When the penalty cap does not bind, settlement risk discounts the value of the execution directly:
+
+$$
+s_{\text{base}}(x) =  p(x)\ S(x)
+$$
+
+A strategy is dominant if it is the best choice regardless of what other solvers submit. This is the dominant strategy in the simplified mechanism where the reward cap does not bind and fairness filtering is ignored, and it is a robust baseline in production. In this setting, that means the solver does not need to predict the reference score or model competitor behaviour to choose the baseline score.
+
+The intuition is that, in this regime, when the relevant caps do not bind and fairness filtering is ignored, changing the reported score only changes whether the solution wins. It does not change the solver’s payoff conditional on that solution winning: lowering the score may lose profitable wins, while raising it may win unprofitable ones.
```
fleupold: These two paragraphs don't fit here (they break the flow between the explanation of the two cases). If you want to still give the intuition, I'd do so at the end in slightly more concise form (also since dominance is not affected by whether the penalty cap binds or not).

```
A strategy is dominant if it is the best choice regardless of what other solvers submit. This is the dominant strategy in the simplified mechanism since the solver does not need to predict the reference score or model competitor behaviour to choose their score.

The intuition is that - given the reward mechanism's second price characteristics - changing the reported score only changes whether the solution wins. It does not change the solver’s payoff. Lowering the score may lose profitable wins, while raising it may win unprofitable ones.
```

docs/cow-protocol/tutorials/solvers/optimal_bidding.md
```diff
@@ -0,0 +1,193 @@
...
+\gamma(x) = S(x)-s(x)+\min(c_l,s(x))
+$$
+
+The solver compares the payoff from settling through adverse slippage with the realised payoff from unsetting. Settling yields approximately $S(x)−s(x)− \gamma$. Unsetting instead incurs a penalty $min(s(x),c_l)$. The slippage tolerance is therefore the point at which the solver is indifferent between these two outcomes.
+
+**Example.** Continuing the previous example, the solver won the 10 ETH order with a score of 95, keeping 5 USDC. Its tolerance is 100 − 95 + 45 = 50 USDC: the 5 USDC it kept, plus the 45 USDC penalty that settling avoids.
+
+| Negative slippage | If the solver settles | If it does not settle | Decision      |
+| ----------------: | --------------------: | --------------------: | ------------- |
+|                30 |          5 − 30 = −25 |                   −45 | Settle        |
```
fleupold: add the line for 50 indifferent (correct tolerance)

docs/cow-protocol/tutorials/solvers/optimal_bidding.md
```diff
@@ -0,0 +1,193 @@
...
+
+Two operational implications:
+
+**Submit individual-pair fallbacks.** If a solver has a batched solution covering multiple pairs, also submit the underlying individual-pair solutions. A high-scoring batched solution that fails fairness is worth zero. The individual-pair fallbacks ensure the solver still wins the pairs it can.
+
+**Verify batched solutions pass the filter.** Compute the per-pair reference outcome from the best individual-pair solutions in the auction and check that the batched solution delivers at least that much on every directed pair it touches.
+
+Individual-vs-batch submission can also be strategic: strong individual-pair bids raise the fairness benchmark and can exclude competitors’ batched solutions, and vice versa. The baseline strategy in this guide does not rely on strategically changing scores to affect the fairness filter, and solvers are not advised to use the fairness filter as a strategic target. Instead, solvers should compute honest individual-pair solutions, submit batched solutions only when they create additional value, and verify that those batched solutions pass the fairness filter.
```
fleupold: Not sure this sentence is useful (it might give them weird ideas)

docs/cow-protocol/tutorials/solvers/optimal_bidding.md
```diff
@@ -0,0 +1,193 @@
...
+
+For each candidate solution $x$:
+
+1. Estimate $S(x)$, $p(x)$.
+2. Compute the baseline score $s_{\text{base}}(x) = \max\!\left( p(x)\ S(x) ,\ \ S(x) - \tfrac{1 - p(x)}{p(x)} \big(c_l\big) \right)$
+3. Set the settlement slippage tolerance using $\gamma(x) = S(x)-s(x)+\min(c_l,s(x))$
+4. For batched solutions, verify the fairness filter passes against the per-pair reference outcome.
+5. For multiple orders on the same directed pair, combine them into one solution.
+6. Submit individual-pair fallbacks alongside batched solutions.
+7. Submit every viable candidate.
```
fleupold: again not clear what viable meanse.

fleupold (commented):
Looking pretty good overall. As a general note, docs PR benefit from putting every sentence into its own line. It makes comments and later edits easier to review.

Also, please let's get rid of the term "baseline" throughout the whole doc (and use recommended or something like that).

docs/cow-protocol/tutorials/solvers/optimal_bidding.md
```diff
@@ -0,0 +1,193 @@
...
+|                           85 |                          +5 |                       +9 | Wins, +9   | Wins, +9 | Wins, +9   |
+|                           93 |                          −3 |                     +1.8 | Wins, +1.8 | Loses, 0 | Wins, +1.8 |
+|                           97 |                          −7 |                     −1.8 | Wins, −1.8 | Loses, 0 | Loses, 0   |
+|                          105 |                         −15 |                       −9 | Loses, 0   | Loses, 0 | Loses, 0   |
+
+Bidding 100 wins an auction that loses money in expectation. Bidding 90 misses a profitable one. Bidding 95 takes every profitable win and skips every loss.
+
+## Slippage and unsuccessful settlements
+
+Solvers can choose not to settle a selected solution, for example, if the desired routing would return significantly fewer tokens than expected. One way to implement this is to set a slippage tolerance $\gamma$: the maximum negative slippage the solver is willing to absorb between bidding and settlement. If on-chain slippage would exceeds $-\gamma$, the settlement does not happen.
```
fleupold: I think the sign on gamma is wrong (if slippage excceeds \gamma not -\gamma

docs/cow-protocol/tutorials/solvers/optimal_bidding.md
Architsharma7: Remove the second paragraph, first tells how winners are chosen, auction picks whichever combination of solutions has the highest total reported score, so it's somewhat relevant here.

Architsharma7 (commented):
Updated based on the reviews.

